PR #145, opened 2018-11-12
Change "Cretonne" to "Cranelift"

sunfishcode:
Also, "cton" to "clif", for completeness.


bjorn3:
Thanks! I forgot to do this when cretonne got renamed to cranelift.

Issue #166, opened 2018-11-17
DWARF support


bjorn3:
Blocked on m4b/faerie#51, gimli-rs/gimli#340 and CraneStation/cranelift#426

* [ ] Unmangled symbols names (only support from `cranelift-faerie` needed)
* [ ] Arguments and locals
* [ ] Line info
* [ ] Inlining information

philipc:
Does this cover eh_frame too?

bjorn3:
Added

bjorn3:
From https://github.com/m4b/faerie/pull/51#issuecomment-443930827 by @philipc:

> https://github.com/philipc/rustc_codegen_cranelift/tree/debuginfo
> https://github.com/philipc/cranelift/tree/debug
>
> Both still WIP. The rustc support will be better once it's using more of rustc_codegen_ssa (eg it has set_debug_loc() calls in all the right places). I'm currently working on .debug_line support in gimli.

I don't see the `.debug_info` contents. :(

(please note that I had to rebase both branches to get it working with the latest nightly)

<details>

```
$ llvm-dwarfdump-3.8 target/out/mini_core_hello_world
target/out/mini_core_hello_world:       file format ELF64-x86-64

.debug_abbrev contents:
Abbrev table for offset: 0x00000000
[1] DW_TAG_compile_unit DW_CHILDREN_yes
        DW_AT_producer  DW_FORM_strp
        DW_AT_language  DW_FORM_udata
        DW_AT_name      DW_FORM_strp
        DW_AT_comp_dir  DW_FORM_strp

[2] DW_TAG_subprogram   DW_CHILDREN_no
        DW_AT_linkage_name      DW_FORM_strp
        DW_AT_decl_file DW_FORM_strp
        DW_AT_decl_line DW_FORM_udata
        DW_AT_decl_column       DW_FORM_udata

Abbrev table for offset: 0x0000001b
[1] DW_TAG_compile_unit DW_CHILDREN_yes
        DW_AT_producer  DW_FORM_strp
        DW_AT_language  DW_FORM_udata
        DW_AT_name      DW_FORM_strp
        DW_AT_comp_dir  DW_FORM_strp

[2] DW_TAG_subprogram   DW_CHILDREN_no
        DW_AT_linkage_name      DW_FORM_strp
        DW_AT_decl_file DW_FORM_strp
        DW_AT_decl_line DW_FORM_udata
        DW_AT_decl_column       DW_FORM_udata


[...]

PR #291, opened 2019-01-17
Implement line debuginfo

bjorn3:
cc @philipc because you are the author of a big part of this
cc #166 

blocked on https://github.com/gimli-rs/gimli/pull/362/


philipc:
Unfortunately I had done a [couple more commits](https://github.com/philipc/rustc_codegen_cranelift/compare/philipc:71fd2eb...philipc:9aad1f1) and forgot to push them :( So a bit of duplicated work there. Not sure if you want to go through my changes and see if there is anything you need, or if you want me to just review what you've done.

bjorn3:
There are a few little changes from your branch I want to incorporate in this PR.

philipc:
I think there's two issues here.

First, yes, .debug_ relocations are not needed for MachO, but that is already handled in faerie, as [discussed before](https://github.com/m4b/faerie/pull/55#discussion_r240874573). Of course, it doesn't hurt to omit them here too.

Second, MachO relocations use implicit addends, so you need to write the addend instead of the 0. This will apply to other targets too (eg ELF `EM_386`). faerie doesn't handle this case for us (and I think [this assert in cranelift](https://github.com/CraneStation/cranelift/blob/a7ea916dcd8e2849f9271324ce3814db36fd43e9/lib/faerie/src/backend.rs#L236) is the same problem).

src/debuginfo.rs
```diff
@@ -0,0 +1,553 @@
...
+    }
+
+    fn emit_location(&mut self, tcx: TyCtxt<'a, 'tcx, 'tcx>, entry_id: UnitEntryId, span: Span) {
+        let loc = tcx.sess.source_map().lookup_char_pos(span.lo());
+
+        let unit = self.units.get_mut(self.unit_id);
+        let entry = unit.get_mut(entry_id);
+
+        let file_id = self.strings.add(loc.file.name.to_string());
+        entry.set(gimli::DW_AT_decl_file, AttributeValue::StringRef(file_id));
```
philipc: This needs to be `AttributeValue::FileIndex`.

src/debuginfo.rs
```diff
@@ -0,0 +1,553 @@
...
+                debug_ranges.0.writer.into_vec(),
+            )
+            .unwrap();
+        artifact
+            .declare_with(
+                SectionId::DebugRngLists.name(),
+                Decl::DebugSection,
+                debug_rnglists.0.writer.into_vec(),
+            )
+            .unwrap();
```
philipc: We should probably only emit this if it isn't empty.

src/debuginfo.rs
```diff
@@ -0,0 +1,553 @@
...
+        &mut self,
+        tcx: TyCtxt,
+        //module: &mut Module<impl Backend>,
+        code_size: u32,
+        context: &Context,
+        isa: &cranelift::codegen::isa::TargetIsa,
+        source_info_set: &indexmap::IndexSet<SourceInfo>,
+    ) {
+        let unit = self.debug_context.units.get_mut(self.debug_context.unit_id);
+        // FIXME: add to appropriate scope intead of root
```
philipc:
```suggestion
```

src/debuginfo.rs
```diff
@@ -0,0 +1,553 @@
...
+        let entry = unit.get_mut(self.entry_id);
+        let mut size_array = [0; 8];
+        target_endian(tcx).write_u64(&mut size_array, code_size as u64);
+        entry.set(gimli::DW_AT_high_pc, AttributeValue::Data8(size_array));
```
philipc: Use `AttributeValue::Udata` instead. If this really needs to be `DW_FORM_data8` then we should have a better gimli API for allowing that instead of doing what you have.

philipc: I'll do a gimli PR that handles the endian conversion there.

src/debuginfo.rs
bjorn3: I do want to add it to the appropriate `DW_TAG_namespace`

src/debuginfo.rs
bjorn3: Gcc emitted `DW_FORM_data8` here.

bjorn3:
I am going to merge. I will try to fix the macOS problems later.

src/debuginfo.rs
philipc: DW_FORM_data1/data2/data4/data8/udata are interchangeable for unsigned constants, and udata will nearly always be smaller in this case.

src/debuginfo.rs
bjorn3: Oops, accidentally duplicated it and though this was the other one.

src/debuginfo.rs
bjorn3: Will use `DW_FORM_udata`.

PR #338, opened 2019-02-09
Update gimli

philipc:
These are the changes to go with https://github.com/gimli-rs/gimli/pull/382.

Let me know if there are any changes you disagree with, or if you have any ideas for further improvements.

### DIFF
--- a/src/debuginfo.rs
+++ b/src/debuginfo.rs
@@ -7,10 +7,9 @@ use std::marker::PhantomData;
 use syntax::source_map::FileName;
 
 use gimli::write::{
-    Address, AttributeValue, CompilationUnit, DebugAbbrev, DebugInfo, DebugLine, DebugRanges,
-    DebugRngLists, DebugStr, EndianVec, LineProgram, LineProgramId, LineProgramTable, Range,
-    RangeList, Result, SectionId, StringTable, UnitEntryId, UnitId, UnitTable,
-    Writer, FileId, LineStringTable, DebugLineStr, LineString,
+    Address, AttributeValue, DwarfUnit, EndianVec, LineProgram, Range,
+    RangeList, Result, SectionId, UnitEntryId,
+    Writer, FileId, LineStringTable, LineString, Sections,
 };
 use gimli::{Encoding, Format, RunTimeEndian};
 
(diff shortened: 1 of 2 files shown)


src/debuginfo.rs
```diff
@@ -51,29 +50,36 @@ fn line_program_add_file(line_program: &mut LineProgram, line_strings: &mut Line
...
-    line_programs: LineProgramTable,
-
-    // Side tables
-    strings: StringTable,
-    line_strings: LineStringTable,
-
-    // Global ids
-    unit_id: UnitId,
-    global_line_program: LineProgramId,
+    dwarf: DwarfUnit,
```
bjorn3: :+1:

Could you please remove the two comments above?

src/debuginfo.rs
```diff
@@ -183,73 +175,37 @@ impl<'a, 'tcx: 'a> DebugContext<'tcx> {
...
-        let debug_ranges_not_empty = !debug_ranges.0.writer.slice().is_empty();
-        if debug_ranges_not_empty {
-            decl_section!(DebugRanges = debug_ranges);
-        }
-
-        let debug_rnglists_not_empty = !debug_rnglists.0.writer.slice().is_empty();
-        if debug_rnglists_not_empty {
-            decl_section!(DebugRngLists = debug_rnglists);
-        }
+        let _: Result<()> = sections.for_each_mut(|id, section| {
```
bjorn3: This is so much cleaner!

src/debuginfo.rs
```diff
@@ -98,45 +106,34 @@ impl<'a, 'tcx: 'a> DebugContext<'tcx> {
...
         };
 
-        let mut strings = StringTable::default();
-        let mut line_strings = LineStringTable::default();
-
-        let mut units = UnitTable::default();
-        let mut line_programs = LineProgramTable::default();
-
-        let global_line_program = line_programs.add(LineProgram::new(
+        let line_program = LineProgram::new(
```
bjorn3: Maybe add a convenience method for these values. (1, 1, -5, 14) These are the ones rustc (or LLVM itself) uses.

philipc: I'm not sure what to do about these yet. Ideally I think these should be selected based on the target arch (at least the first two parameters, but maybe the others as well).

bjorn3:
I will have to update for latest nightly. Will merge later today.

bjorn3:
Thanks!

PR #351, opened 2019-02-18
Don't sign commits from patches in sysroot

milkey-mouse:
I have commit signing enabled by default globally, so whenever I prepare the sysroot it asks me to unlock my GPG key. A signature is unnecessary here anyway, so disable them with `--no-gpg-sign`.


bjorn3:
Thanks.

PR #495, opened 2019-04-24
Use cg clif link

bjorn3:
Blocked on https://github.com/mdsteele/rust-ar/pull/11


bjorn3:
MacOS somehow gives a linker error. I will investigate later this week.

bjorn3:
Turns out `rust.metadata.bin` sometimes got added to the start of `.rlib` files. `ld` gets confused by this. I fixed it by changing a `HashMap` to a `Vec`.

PR #709, opened 2019-09-12
Update rustc and dependencies

bjorn3:
This enables the backtrace feature of libstd, because it is broken without (https://github.com/rust-lang/rust/issues/64410) I hope this doesn't break macOS


bjorn3:
I removed the patch for `object`, but forgot to increase its version number in `Cargo.toml`, so the broken version without https://github.com/gimli-rs/object/pull/109 was used again, causing dylib metadata loading on macOS to fail.

PR #758, opened 2019-10-16
Add object::write as alternative object file writer

bjorn3:
Based on https://github.com/philipc/rustc_codegen_cranelift/commit/3db299698363041b3a6c4182115115031e6796dc by @philipc.

## TODO

* [x] Change driver.rs to use when a feature flag is set.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1,10 +1,13 @@
 use std::collections::HashMap;
 
-use cranelift_module::FuncId;
+use rustc::session::Session;
+
+use cranelift_module::{FuncId, Module};
 
 use faerie::*;
 use object::{SectionKind, RelocationKind, RelocationEncoding};
 use object::write::*;
+use cranelift_faerie::{FaerieBackend, FaerieBuilder, FaerieProduct, FaerieTrapCollection};
 use cranelift_object::*;
 
 use gimli::SectionId;
@@ -73,11 +76,11 @@ pub trait WriteDebugInfo {
     );
 }
 
-impl WriteDebugInfo for Artifact {
+impl WriteDebugInfo for FaerieProduct {
     type SectionId = SectionId;
 
     fn add_debug_section(&mut self, id: SectionId, data: Vec<u8>) -> SectionId {
-        self.declare_with(id.name(), Decl::section(faerie::SectionKind::Debug), data).unwrap();
+        self.artifact.declare_with(id.name(), Decl::section(faerie::SectionKind::Debug), data).unwrap();
         id
     }
 
@@ -89,6 +92,7 @@ impl WriteDebugInfo for Artifact {
         reloc: &DebugReloc,
     ) {
         self
+            .artifact
             .link_with(
                 faerie::Link {
                     from: from.name(),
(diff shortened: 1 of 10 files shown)


philipc:
Sorry I haven't had time to finish this off. Let me know if I can do anything to help debug the osx failure.

bjorn3:
No problem. The macOS failure is not caused by this. It fails on master too: https://github.com/bjorn3/rustc_codegen_cranelift/issues/728. Upstream issue: https://github.com/rust-lang/rust/issues/64872.

bjorn3:
@philipc Do you want to review this PR before I merge?

src/backend.rs
```diff
@@ -0,0 +1,219 @@
...
+            .unwrap();
+    }
+}
+
+impl WriteMetadata for object::write::Object {
+    fn add_rustc_section(&mut self, symbol_name: String, data: Vec<u8>, is_like_osx: bool) {
+        let segment = self.segment_name(object::write::StandardSegment::Data).to_vec();
+        let section_id = self.add_section(segment, b".rustc".to_vec(), object::SectionKind::Data);
+        let offset = self.append_section_data(section_id, &data, 1);
+        // FIXME implement faerie elf backend section custom symbols
```
philipc: What does this comment mean? Just wondering if this is something that can be fixed already.

src/debuginfo.rs
```diff
@@ -56,31 +56,22 @@ fn line_program_add_file(
...
-            DebugRelocName::Section(id) => id.name(),
-            DebugRelocName::Symbol(index) => ctx.symbols.get_index(index).unwrap(),
-        }
-    }
-}
-
 pub struct DebugContext<'tcx> {
     endian: RunTimeEndian,
-    symbols: indexmap::IndexSet<String>,
+    symbols: indexmap::IndexSet<(String, FuncId)>,
```
philipc: I'm not happy with this type for two reasons, but not sure if there is a way to improve it.

-  I tried to avoid unnecessary copies of symbol names for cranelift_object, but that seems to have just made things more complicated because since faerie still needs the copies.

- `FuncId` is already an index itself. However, not sure if we can use that, because maybe we'll need to change this to `FuncOrDataId` in future. At the least, it shouldn't be necessary to hash on the string, so I think `IndexMap<FuncId, String>` could be used.

src/driver.rs
```diff
@@ -147,59 +147,57 @@ fn run_aot(
...
             let tmp_file = tcx
                 .output_filenames(LOCAL_CRATE)
-                .temp_path(OutputType::Object, Some(&artifact.name));
-            let obj = artifact.emit().unwrap();
+                .temp_path(OutputType::Object, Some(&name));
+            let obj = product.emit();
             std::fs::write(&tmp_file, obj).unwrap();
             CompiledModule {
-                name: artifact.name,
+                name: name,
```
philipc:
```suggestion
                name,
```

src/backend.rs
bjorn3: The elf backend of faerie doesnt support `add_symbol`.

src/debuginfo.rs
bjorn3: It seems that `FuncId` has an `as_u32` method, so it is possible to get a `u32` index as required by gimli. This means that using `IndexMap<FuncId, String>` could work. For `FuncIdOrDataId` that would not work unforunately.

src/backend.rs
philipc: Ah ok, so this could be deleted here then? I think Windows will need to add this symbol too.

src/debuginfo.rs
bjorn3: It should be possible to encode if a `FuncIdOrDataId` is a `FuncId` or a `DataId` in the most significant bit.

bjorn3: Oh I see, `IndexMap` returns an index too.

bjorn3:
> I tried testing this for Windows but something unrelated to this PR has changed and I'm getting errors when linking:
>
> ```
> note: LINK : fatal error LNK1181: cannot open input file 'c.lib'
> ```

Thanks for testing.

Disclaimer: I have never used a compiler on Windows.

It seems to me that the linker couldn't find libc. `rustc` without cg_clif works for you right? cg_clif should use the same linker invocation as cg_llvm, as cg_clif uses `rustc_codegen_ssa::back::link::link_binary` just like cg_llvm.

philipc:
> rustc without cg_clif works for you right? cg_clif should use the same linker invocation as cg_llvm, as cg_clif uses `rustc_codegen_ssa::back::link::link_binary` just like cg_llvm.

`rustc` works for normal crates. But `test.sh` doesn't work even if I'm using rustc_codegen_llvm instead of cg_clif. I'll need to compare linker invocations with the older version to see what's changed.

bjorn3:
I think I know. It is `mini_core` which failed, isn't it? That one links to `libc`. I saw that you changed that to `msvcrt` on your windows branch, which I didnt cherry-pick.

philipc:
Thanks, yep that's what it is. I don't even remember doing that change. There's other changes needed too. I'll work on a PR for Windows once this is merged.

PR #839, opened 2019-12-20
Some runtime optimizations

bjorn3:
* Don't force return value of call to stack
* Use load and store offsets instead of iadd_imm

```
Benchmark #1: ./main_before
  Time (mean ± σ):     13.579 s ±  0.186 s    [User: 13.567 s, System: 0.007 s]
  Range (min … max):   13.403 s … 14.002 s    10 runs
 
Benchmark #2: target/x86_64-unknown-linux-gnu/debug/main
  Time (mean ± σ):     12.314 s ±  0.033 s    [User: 12.297 s, System: 0.010 s]
  Range (min … max):   12.243 s … 12.350 s    10 runs
 
Summary
  'target/x86_64-unknown-linux-gnu/debug/main' ran
    1.10 ± 0.02 times faster than './main_before'
```

For comparison LLVM + debug mode:

```
Benchmark #1: target/debug/main
  Time (mean ± σ):      7.929 s ±  0.291 s    [User: 7.906 s, System: 0.013 s]
  Range (min … max):    7.589 s …  8.293 s    10 runs
```

### DIFF
--- a/src/abi/comments.rs
+++ b/src/abi/comments.rs
@@ -94,14 +94,14 @@ pub fn add_local_place_comments<'tcx>(
             align.abi.bytes(),
             align.pref.bytes(),
         )),
-        CPlaceInner::Addr(addr, None) => fx.add_global_comment(format!(
-            "reuse {:5} {:20} {:4}b {}, {}              storage={}",
+        CPlaceInner::Addr(ptr, None) => fx.add_global_comment(format!(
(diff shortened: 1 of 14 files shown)


bjorn3:
The last commit uses `stack_load` and `stack_store` where possible. While Cranelift currently expands them to `stack_addr` + `load`/`store`, it can give a performance improvement once Cranelift uses a single `mov` on x86.

Issue #171, opened 2018-11-20
std::arch SIMD intrinsics


gnzlbg:
Currently the SIMD intrinsics are implemented in `stdsimd` using `link_llvm_intrinsics` to directly call the llvm intrinsics with the appropriate types, and using a handful of "generic" simd intrinsics. 

Is there a way to directly call Cretonne intrinsics?
Is there a `cfg()` macro available to detect whether the codegen backend is LLVM or Cranelift ?

gnzlbg:
@bjorn3 check this out:

https://github.com/rust-lang-nursery/stdsimd/blob/master/coresimd/x86/sse.rs#L1986

This is how `std::arch` calls into the different simd instructions (not all of them, but many of them). The question is, would it be possible to write similar code to target Cranelift ? Or should we move all of these into `platform-intrinsics` and add the abstraction layer at the Rust codegen level ?

> It does have simd types, which can be used with normal instructiosn like iadd

That sounds more like what `packed_simd` does, which uses some of rustc's generic simd intrinsics, e.g., see here: https://github.com/rust-lang-nursery/packed_simd/blob/master/src/codegen/llvm.rs

It might be easier to get `packed_simd` to work with Cranelift than to get `std::arch` working, but note that `packed_simd` also uses `std::arch` intrinsics in many cases to work arounds codegen bugs in LLVM... A cfg macro to detect the backend would be needed here to detect Cranelift and remove these workarounds, but we might have to potentially add other workarounds for Cranelift, by "somehow" calling Cranelift SIMD operations.

bjorn3:
`cg_clif` currently puts non-primitives in stackslots which would kill simd performance.

bjorn3:
> The question is, would it be possible to write similar code to target Cranelift ?

Not without changes to `cg_clif` to intercept those intrinsics.

> Even if this work is not there yet, I think work to "prepare" std::arch and packed_simdfor Cranelift can already start, and ideally using such a cfg macro we would get one single std::arch intrinsic first, write down the process to "support Cranelift", and try to get people involved, mentor them, etc.

+1

gnzlbg:
What would be desirable to start preparing `stdsimd` and `packed_simd`, is to clarify how the result should look like in those crates. cc @eddyb @sunfishcode 

[...]

PR #842, opened 2019-12-23
Codegen simd intrinsics as simd clif instructions

bjorn3:
Simd values are still stored on the stack, so this doesn't improve performance much yet.

cc #171


bjorn3:
This spuriously hits https://github.com/bytecodealliance/cranelift/issues/1311. Going to cherry pick some comments and keeping the rest until https://github.com/bytecodealliance/cranelift/issues/1311 is fixed.

bjorn3:
I think the error was because I pushed a bad commit between the moment travis job of the previous commit began and the moment the macOS builders checked out this PR.

PR #866, opened 2020-01-15
Store nightly snapshot in rust-toolchain file, and use it in cargo.sh

eminence:
The rust-toolchain file can also store an exact nightly snapshot
(instead of just "nightly"), so we can store whatever snapshot that
rustc_codegen_cranelift is known to work with.

This also lets us add a new feature to `cargo.sh` to let it use the
exact same nightly snapshot as cg_clif.  If there's a nightly compiler
mismatch, you get a confusing error message like:

    error: couldn't load codegen backend "librustc_codegen_cranelift.so":
    "librustc_driver-681e2b4f66c73d3e.so: cannot open
    shared object file: No such file or directory"

So doing this automatically in cargo.sh is useful.

One consequence of this change is that you'll have to remember to keep `rust-toolchain` up-to-date.

### DIFF
--- a/cargo.sh
+++ b/cargo.sh
@@ -6,9 +6,13 @@ fi
 
 pushd $(dirname "$0") >/dev/null
 source config.sh
+
+# read nightly compiler from rust-toolchain file
+TOOLCHAIN=$(cat rust-toolchain)
+
 popd >/dev/null
 
 cmd=$1
 shift
 
-cargo $cmd --target $TARGET_TRIPLE $@
+cargo +${TOOLCHAIN} $cmd --target $TARGET_TRIPLE $@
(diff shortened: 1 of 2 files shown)


bjorn3:
> One consequence of this change is that you'll have to remember to keep rust-toolchain up-to-date.

This is the main reason I didn't do this. Another is that I will have to manually remove the old nightly each time I update. I think I will make a rustup script to automate this and committing the changes.

eminence:
If you didn't want to modify `rust-toolchain`, then that rustup script you're talking about could also be used by other people to make sure they are using a compatible nightly snapshot.

bjorn3:
What do you think about https://github.com/bjorn3/rustc_codegen_cranelift/tree/pin_toolchain?

eminence:
Thanks for merging this.  I was able to do a `git pull` and have `./prepare.sh` automatically download the correct snapshot, which is pretty neat

PR #918, opened 2020-03-11
Implement incremental caching of object files

### LINKED ISSUE #760: Cache compiled codegen units in incremental cache
It should be behind a feature flag, to make it possible to use incremental mode while working on cg_clif.

bjorn3:
It is only enabled when the `CG_CLIF_INCR_CACHE` env var is set to prevent accidentally caching object files in the incremental cache while modifying cg_clif. Otherwise it would be necessary to completely disable the incremental cache during development, which makes the sysroot build much slower.

Fixes #760

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -18,6 +18,7 @@ $ ./test.sh --release
 ### Cargo
 
 ```bash
+$ export CG_CLIF_INCR_CACHE=1 # Enable caching of object files in the incremental cache
 $ CHANNEL="release" $cg_clif_dir/cargo.sh run
 ```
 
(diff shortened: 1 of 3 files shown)


bjorn3:
Currently running the rustc-perf suite. Hopefully this makes the incremental benches mostly green too, like the non-incremental benches.

bjorn3:
I think something went wrong with the incremental benches. The results for cg_clif are still as horrible as before.

bjorn3:
I think it is now correctly benchmarking. I moved the `CG_CLIF_INCR_CACHE=1` from the `./target/release/collector` invocation to the source of the collector.

bjorn3:
Posted results in https://github.com/bjorn3/rustc_codegen_cranelift/issues/878#issuecomment-597871730. tl;dr: most incremental benchmarks are now faster with cg_clif than with cg_llvm, but there are still regressions.

PR #951, opened 2020-03-31
Simd improvements

bjorn3:


bjorn3:
```
failures:

---- num::bignum::test_mul_pow2_overflow_2 stdout ----
---- num::bignum::test_mul_pow2_overflow_2 stderr ----
Unexpected error: child process exited with signal 4

failures:
    num::bignum::test_mul_pow2_overflow_2
```

Bisected to de855002c8e3583ff69bf383f4bbae17f8c7a3d4. Omitting it for now.

bjorn3:
There is now a macOS only `leading_zeros()` problem caused by using the `haswell` preset. Debugging it is hard because I don't have access to a mac right now, so I have to use CI.

bjorn3:
Cherry-picked the changed not related to SIMD onto master.

bjorn3:
Found the problem: The macOS CI uses an "Intel(R) Xeon(R) CPU E5-1650 v2 @ 3.50GHz" This is a Ivy Bridge EP, which means that lzcnt is not yet supported.

bjorn3:
Switched the target cpu from haswell to nehalem, which is good enough for SIMD support.

PR #972, opened 2020-04-18
Support file hashes in .debug_line

### LINKED ISSUE #936: Add file hash to .debug_line
The file hash can be used by a debugger to detect when the source file has changed after compilation.

This can be implemented by passing a `FileInfo` the right MD5 hash to [`LineProgram::new`](https://github.com/bjorn3/rustc_codegen_cranelift/blob/b113e88ddb399f92f0d044caa35e668d2499bdd9/src/debuginfo/mod.rs#L65) and `line_program.add_file` ([here](https://github.com/bjorn3/rustc_codegen_cranelift/blob/b113e88ddb399f92f0d044caa35e668d2499bdd9/src/debuginfo/line_info.rs#L60) and [here](https://github.com/bjorn3/rustc_codegen_cranelift/blob/b113e88ddb399f92f0d044caa35e668d2499bdd9/src/debuginfo/line_info.rs#L70)) and setting `line_program.file_has_md5` to true. Also the DWARF version on platforms other than macOS needs to be changed to 5.

Blocked on rust-lang/rust#69718, which implements this for cg_llvm and adds a new `src_hash` field to `SourceFile` storing the MD5 hash.

l4l:
resolves #936

### DIFF
--- a/src/debuginfo/line_info.rs
+++ b/src/debuginfo/line_info.rs
@@ -1,5 +1,4 @@
 use std::ffi::OsStr;
-use std::convert::TryFrom;
 use std::path::{Component, Path};
 
 use crate::prelude::*;
@@ -42,24 +41,18 @@ pub(crate) const MD5_LEN: usize = 16;
 pub struct FileHash([u8; MD5_LEN]);
 
 impl FileHash {
-    pub fn inner(self) -> [u8; MD5_LEN] {
-        self.0
-    }
-}
-
-pub struct UnsupportedHashType;
-
-impl TryFrom<SourceFileHash> for FileHash {
-    type Error = UnsupportedHashType;
-
-    fn try_from(hash: SourceFileHash) -> Result<Self, Self::Error> {
+    pub fn from_source_hash(hash: SourceFileHash) -> Option<Self> {
         if hash.kind == SourceFileHashAlgorithm::Md5 {
             let mut buf = [0u8; MD5_LEN];
             buf.copy_from_slice(hash.hash_bytes());
-            Ok(Self(buf))
+            Some(Self(buf))
         } else {
-            Err(UnsupportedHashType)
+            None
         }
+    }
+
+    pub fn inner(self) -> [u8; MD5_LEN] {
+        self.0
     }
 }
 
@@ -86,9 +79,9 @@ fn line_program_add_file(
                 line_strings,
             );
 
-            let file_hash = FileHash::try_from(file.src_hash);
+            let file_hash = FileHash::from_source_hash(file.src_hash);
 
-            line_program.file_has_md5 = file_hash.is_ok();
+            line_program.file_has_md5 = file_hash.is_some();
             line_program.add_file(file_name, dir_id, Some(FileInfo {
                 timestamp: 0,
                 size: 0,
--- a/src/debuginfo/line_info.rs
+++ b/src/debuginfo/line_info.rs
@@ -1,5 +1,4 @@
 use std::ffi::OsStr;
-use std::time::SystemTime;
 use std::path::{Component, Path};
 
 use crate::prelude::*;
@@ -67,14 +66,10 @@ fn line_program_add_file(
                     buf
                 });
 
-            line_program.file_has_timestamp = true;
             line_program.file_has_md5 = md5.is_some();
 
             line_program.add_file(file_name, dir_id, Some(FileInfo {
-                timestamp: SystemTime::now()
-                    .duration_since(SystemTime::UNIX_EPOCH)
-                    .map(|t| t.as_secs())
-                    .unwrap_or(0),
+                timestamp: 0,
                 size: 0,
                 md5: md5.unwrap_or_default(),
             }))
--- a/src/debuginfo/line_info.rs
+++ b/src/debuginfo/line_info.rs
@@ -1,14 +1,15 @@
 use std::ffi::OsStr;
+use std::time::SystemTime;
 use std::path::{Component, Path};
 
 use crate::prelude::*;
 
-use rustc_span::{FileName, SourceFileAndLine, Pos};
+use rustc_span::{FileName, SourceFile, SourceFileAndLine, Pos, SourceFileHash, SourceFileHashAlgorithm};
 
 use cranelift_codegen::binemit::CodeOffset;
 
 use gimli::write::{
-    Address, AttributeValue, FileId, LineProgram, LineString, LineStringTable, UnitEntryId,
+    Address, AttributeValue, FileId, LineProgram, LineString, FileInfo, LineStringTable, UnitEntryId,
 };
 
 // OPTIMIZATION: It is cheaper to do this in one pass than using `.parent()` and `.file_name()`.
@@ -38,9 +39,9 @@ fn osstr_as_utf8_bytes(path: &OsStr) -> &[u8] {
 fn line_program_add_file(
     line_program: &mut LineProgram,
     line_strings: &mut LineStringTable,
-    file: &FileName,
+    file: &SourceFile,
 ) -> FileId {
-    match file {
+    match &file.name {
         FileName::Real(path) => {
             let (dir_path, file_name) = split_path_dir_and_file(path);
             let dir_name = osstr_as_utf8_bytes(dir_path.as_os_str());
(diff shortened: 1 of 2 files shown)


src/debuginfo/line_info.rs
```diff
@@ -57,13 +58,32 @@ fn line_program_add_file(
...
+                });
+
+            line_program.file_has_timestamp = true;
+            line_program.file_has_md5 = md5.is_some();
+
+            line_program.add_file(file_name, dir_id, Some(FileInfo {
+                timestamp: SystemTime::now()
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .map(|t| t.as_secs())
+                    .unwrap_or(0),
```
bjorn3, line 77: I would prefer to not include a timestamp to help with reproducible builds.

src/debuginfo/line_info.rs
```diff
@@ -57,13 +58,32 @@ fn line_program_add_file(
                 line_program.encoding(),
                 line_strings,
             );
-            line_program.add_file(file_name, dir_id, None)
+
+            let md5 = Some(file.src_hash)
+                .filter(|h| matches!(h, SourceFileHash { kind: SourceFileHashAlgorithm::Md5, .. }))
```
bjorn3, line 63:
```suggestion
                .filter(|h| h.kind == SourceFileHashAlgorithm::Md5)
```

src/debuginfo/mod.rs
```diff
@@ -42,7 +48,8 @@ impl<'tcx> DebugContext<'tcx> {
             format: Format::Dwarf32,
             // TODO: this should be configurable
             // macOS doesn't seem to support DWARF > 3
-            version: 3,
+            // 5 version is required for md5 file hash
+            version: 5,
```
bjorn3, line 52:
```suggestion
            version: if tcx.sess.target.target.options.is_like_osx {
                3
            } else {
                5
            },
```

src/debuginfo/mod.rs
```diff
@@ -52,18 +59,41 @@ impl<'tcx> DebugContext<'tcx> {
...
+                    .source_map()
+                    .get_source_file(&FileName::Real(path))
+                    .map(|f| f.src_hash)
+                    .filter(|h| matches!(h, SourceFileHash { kind: SourceFileHashAlgorithm::Md5, .. }))
+                    .map(|h| {
+                        let mut buf = [0u8; MD5_LEN];
+                        buf.copy_from_slice(h.hash_bytes());
+                        buf
+                    });
+                (name, hash)
```
bjorn3, line 75: It would be nice to deduplicate this logic to create `FileInfo`. It should probably be in `line_info.rs`

src/debuginfo/line_info.rs
```diff
@@ -57,13 +58,32 @@ fn line_program_add_file(
...
+                .map(|h| {
+                    let mut buf = [0u8; super::MD5_LEN];
+                    buf.copy_from_slice(h.hash_bytes());
+                    buf
+                });
+
+            line_program.file_has_timestamp = true;
+            line_program.file_has_md5 = md5.is_some();
```
bjorn3, line 71: It may be possible for files coming from different crates (due to for example macros) to have different hash algorithms. Does always setting this to `true` work? Or will gdb or lldb think an MD5 of 0 means that the file should actually hash to 0?

l4l: [Seems](https://docs.rs/gimli/0.20.0/gimli/write/struct.FileInfo.html#structfield.md5) that `true` here means zeroed md5 considered as a hash.

What exactly do you mean for "files coming from different crates"? I think `SourceFile` should have some info about its crate, so even same files from different crates are added separately here, isn't it?

bjorn3: I think if one crate uses no hash and it is imported in another crate using md5 as hash, that the files from the imported crate will still get no hash, while the files of the other crate will get an md5 hash.

src/debuginfo/line_info.rs
```diff
@@ -35,12 +36,39 @@ fn osstr_as_utf8_bytes(path: &OsStr) -> &[u8] {
...
+    fn try_from(hash: SourceFileHash) -> Result<Self, Self::Error> {
+        if hash.kind == SourceFileHashAlgorithm::Md5 {
+            let mut buf = [0u8; MD5_LEN];
+            buf.copy_from_slice(hash.hash_bytes());
+            Ok(Self(buf))
+        } else {
+            Err(UnsupportedHashType)
+        }
+    }
+}
```
bjorn3, line 64: Maybe add a method to create an `Option<FileInfo>` from `SourceFileHash` instead. Where `None` means unsupported hash type. You could then directly pass it to `add_file`.

src/debuginfo/line_info.rs
bjorn3: If you pass `None` for `file_info` to `add_file`, gimli will write a `FileInfo` with `md5` set to 0.

src/debuginfo/line_info.rs
```diff
@@ -57,13 +85,21 @@ fn line_program_add_file(
...
         // FIXME give more appropriate file names
-        _ => {
+        filename => {
             let dir_id = line_program.default_directory();
             let dummy_file_name = LineString::new(
-                file.to_string().into_bytes(),
+                filename.to_string().into_bytes(),
                 line_program.encoding(),
                 line_strings,
             );
```
l4l, line 105: Should this branch also provide `FileInfo`?

src/debuginfo/line_info.rs
l4l: What are these cases, when that hash is not available? I'm not sure that can even possible

src/debuginfo/line_info.rs
bjorn3: No, this is the fallback for when there is no file on the disk, because it was for example a macro expansion, or read from stdin. (Fun fact: `echo "fn main() {} | rustc -` works)

src/debuginfo/line_info.rs
bjorn3: I actually meant a standalone function to create `Option<gimli::write::FileInfo>`, not `Option<FileHash>`.

l4l: Ah, got it, thanks for the explanation

src/debuginfo/line_info.rs
bjorn3: You can switch to an SHA-1 hash using `-Zsrc-hash-algorithm=sha1`.

l4l: Quiet weird scenario, if I understood correctly. So the complete case is:

- compile crate Y with defaults;
- compile crate X (whose depends on Y) with sha1 src-hash;

Now if include both crates there will be 2 X-es, but only one with a debug file hashes, right?

bjorn3: I was referring to:

```rust
// foo/lib.rs (compile with defaults)
pub fn foo_fn<T>() {}

// bar/lib.rs (compile with sha1 src-hash)
pub fn call_foo_fn() {
    foo_fn::<()>();
}
```

In crate bar, the debuginfo for `foo_fn::<()>` will refer to `foo/lib.rs` with md5 hash, while the debuginfo of `call_foo_fn` will refer to `bar/lib.rs` with sha1 hash.

philipc: > If you pass None for file_info to add_file, gimli will write a FileInfo with md5 set to 0.

That seems like a bug in gimli. LLVM only emits the MD5 field if all fields have MD5 info. DWARF does not support only having MD5 fields for some of the files.

src/debuginfo/line_info.rs
```diff
@@ -57,13 +73,17 @@ fn line_program_add_file(
...
-            line_program.add_file(file_name, dir_id, None)
+
+            let info = make_file_info(file.src_hash);
+
+            line_program.file_has_md5 = info.is_some();
```
bjorn3, line 79:
```suggestion
            line_program.file_has_md5 &= info.is_some();
```

According to https://github.com/bjorn3/rustc_codegen_cranelift/pull/972#discussion_r410781643 it is either every file has an MD5 or none.

l4l: Thanks, sorry for a delay

bjorn3:
Thanks for the PR! Unfortunately I just noticed that my gdb and lldb don't yet support DWARF version 5. I have switched this PR to version 4 for now, which means that no md5 hash is currently emitted. Once my gdb and lldb support it, I will switch back to DWARF version 5. I will merge this PR anyway, as it does almost all of the work necessary to support md5 checksums.

PR #1021, opened 2020-05-29
Mark blocks that call cold funs as cold

osa1:
@bjorn3 I couldn't test this as I'm not getting any clif output even when I don't pass `--release` to `test.sh`.

I'm also not sure about how to format this code, it seems like the toolchain we use don't have rustfmt.

### DIFF
--- a/src/abi/mod.rs
+++ b/src/abi/mod.rs
@@ -4,6 +4,7 @@ mod pass_mode;
 mod returning;
 
 use rustc_target::spec::abi::Abi;
+use rustc_middle::middle::codegen_fn_attrs::CodegenFnAttrFlags;
 
 use cranelift_codegen::ir::AbiParam;
 
@@ -431,6 +432,7 @@ pub(crate) fn codegen_fn_prelude<'tcx>(
 pub(crate) fn codegen_terminator_call<'tcx>(
     fx: &mut FunctionCx<'_, 'tcx, impl Backend>,
     span: Span,
+    current_block: Block,
     func: &Operand<'tcx>,
     args: &[Operand<'tcx>],
     destination: Option<(Place<'tcx>, BasicBlock)>,
@@ -439,8 +441,6 @@ pub(crate) fn codegen_terminator_call<'tcx>(
     let fn_sig = fx
         .tcx
         .normalize_erasing_late_bound_regions(ParamEnv::reveal_all(), &fn_ty.fn_sig(fx.tcx));
-
-    // FIXME mark the current block as cold when calling a `#[cold]` function.
 
     let destination = destination.map(|(place, bb)| (trans_place(fx, place), bb));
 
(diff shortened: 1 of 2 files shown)


osa1:
I confirmed that changes order of blocks. My test program:

```rust
fn main() {
    let mut s = String::new();
    ::std::io::stdin().read_line(&mut s).unwrap();

    if str::parse::<i64>(&s).unwrap().is_positive() {
        a();
    } else {
        b();
    }
}

#[inline(never)]
#[cold]
fn a() {
    println!("a called");
}

#[inline(never)]
fn b() {
    println!("b called");
}
```

Build with master this is the optimized clif I get: (only the relevant bits)

```
@0012                               brz v26, block11
@0012                               jump block12

                                block11:
@0012                               nop 
; 
; _16 = const b()
@0013                               call fn9()
@0013                               jump block14

                                block12:
@0013                               nop 
; 
; _15 = const a()
@0014                               call fn10()
@0014                               jump block13
...
```

With this patch target of the conditional jump is moved to the end of the file:

```
@0012 [RexOp1tjccb#74]              brz v26, block11
@0012 [Op1jmpb#eb]                  jump block12

                                block11:
@0012 [-]                           nop 
; 
; _16 = const b()
@0013 [Op1call_id#e8]               call fn9()
@0013 [Op1jmpb#eb]                  jump block14

                                block13:
...


... at the end of the file ...
                                block12:
@0013 [-]                           nop 
; 
; _15 = const a()
@0014 [Op1call_id#e8]               call fn10()
@0014 [Op1jmpb#eb]                  jump block13
}
```

One thing that I'm not sure about is why we we move the "else" branch of `brz`, that is, the block that we jump to when `is_positive()` returns false is moved. I think the other block should've been moved? Maybe I'm reading the clif output wrong? Or maybe `true` is represented as `0`?

bjorn3:
> I'm also not sure about how to format this code, it seems like the toolchain we use don't have rustfmt.

I don't really use rustfmt. I try to follow it's style as much as possible, but I always forget to actually run rustfmt.

> One thing that I'm not sure about is why we we move the "else" branch of brz, that is, the block that we jump to when is_positive() returns false is moved. I think the other block should've been moved? Maybe I'm reading the clif output wrong? Or maybe true is represented as 0?

```
    fn9 = colocated u0:14 sig9 ; Instance { def: Item(DefId(0:8 ~ std_example[317d]::b[0])), substs: [] }
    fn10 = colocated u0:7 sig10 ; Instance { def: Item(DefId(0:7 ~ std_example[317d]::a[0])), substs: [] }
```

Surprisingly `fn9` is `b` and `fn10` is `a`.

src/abi/mod.rs
```diff
@@ -442,6 +444,7 @@ pub(crate) fn codegen_terminator_call<'tcx>(
 
     // FIXME mark the current block as cold when calling a `#[cold]` function.
```
bjorn3, line 445: Please remove this FIXME and ensure that there is exactly one blank line between the `normalize_erasing_late_bound_regions` above and the `let destination` below.

osa1: Done.

bjorn3:
Thanks a lot!

PR #1024, opened 2020-05-30
Simplify target_arch conditional in codegen_crate

osa1:
I think using `cfg!` here is easier to read, no need for two conditional items.


bjorn3:
The `jit` module doesn't exist at all on `wasm32` https://github.com/bjorn3/rustc_codegen_cranelift/blob/1e70c51f60abc544d9a325c77474190f1719e560/src/driver/mod.rs#L9-L10 because `cranelift-simplejit` fails to compile for `wasm32` and as such had to be disabled at https://github.com/bjorn3/rustc_codegen_cranelift/blob/1e70c51f60abc544d9a325c77474190f1719e560/Cargo.toml#L44-L46

osa1:
Ahh, right. OK in that case this can be closed.

PR #1064, opened 2020-07-10
Basic inline asm support

bjorn3:
You need to use fixed registers. Register classes are not yet supported.

### DIFF
--- a/src/inline_asm.rs
+++ b/src/inline_asm.rs
@@ -1,0 +1,220 @@
+use crate::prelude::*;
+
+use std::fmt::Write;
+
+use rustc_ast::ast::{InlineAsmTemplatePiece, InlineAsmOptions};
+use rustc_middle::mir::InlineAsmOperand;
+use rustc_target::asm::*;
+
+pub(crate) fn codegen_inline_asm<'tcx>(
+    fx: &mut FunctionCx<'_, 'tcx, impl Backend>,
+    _span: Span,
+    template: &[InlineAsmTemplatePiece],
+    operands: &[InlineAsmOperand<'tcx>],
+    options: InlineAsmOptions,
+) {
+    // FIXME add .eh_frame unwind info directives
+
+    if template.is_empty() {
+        // Black box
+        return;
+    }
+
+    let mut slot_size = Size::from_bytes(0);
+    let mut clobbered_regs = Vec::new();
+    let mut inputs = Vec::new();
+    let mut outputs = Vec::new();
+
+    let mut new_slot = |reg_class: InlineAsmRegClass| {
+        let reg_size = reg_class
+            .supported_types(InlineAsmArch::X86_64)
+            .iter()
+            .map(|(ty, _)| ty.size())
+            .max()
+            .unwrap();
+        let align = rustc_target::abi::Align::from_bytes(reg_size.bytes()).unwrap();
+        slot_size = slot_size.align_to(align);
+        let offset = slot_size;
+        slot_size += reg_size;
+        offset
+    };
+
+    // FIXME overlap input and output slots to save stack space
+    for operand in operands {
+        match *operand {
+            InlineAsmOperand::In { reg, ref value } => {
+                let reg = expect_reg(reg);
+                clobbered_regs.push((reg, new_slot(reg.reg_class())));
+                inputs.push((reg, new_slot(reg.reg_class()), crate::base::trans_operand(fx, value).load_scalar(fx)));
+            }
+            InlineAsmOperand::Out { reg, late: _, place } => {
+                let reg = expect_reg(reg);
+                clobbered_regs.push((reg, new_slot(reg.reg_class())));
+                if let Some(place) = place {
+                    outputs.push((reg, new_slot(reg.reg_class()), crate::base::trans_place(fx, place)));
+                }
+            }
+            InlineAsmOperand::InOut { reg, late: _, ref in_value, out_place } => {
+                let reg = expect_reg(reg);
+                clobbered_regs.push((reg, new_slot(reg.reg_class())));
+                inputs.push((reg, new_slot(reg.reg_class()), crate::base::trans_operand(fx, in_value).load_scalar(fx)));
+                if let Some(out_place) = out_place {
+                    outputs.push((reg, new_slot(reg.reg_class()), crate::base::trans_place(fx, out_place)));
+                }
+            }
+            InlineAsmOperand::Const { value: _ } => todo!(),
+            InlineAsmOperand::SymFn { value: _ } => todo!(),
+            InlineAsmOperand::SymStatic { def_id: _ } => todo!(),
+        }
+    }
+
+    let asm_name = format!("{}__inline_asm_{}", fx.tcx.symbol_name(fx.instance).name, /*FIXME*/0);
+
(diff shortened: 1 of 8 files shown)


src/inline_asm.rs
```diff
@@ -0,0 +1,227 @@
...
+    let mut generated_asm = String::new();
+    writeln!(generated_asm, ".globl {}", asm_name).unwrap();
+    writeln!(generated_asm, "{}:", asm_name).unwrap();
+
+    generated_asm.push_str(".intel_syntax noprefix\n");
+    generated_asm.push_str("    push rbp\n");
+    generated_asm.push_str("    mov rbp,rdi\n");
+    if !options.contains(InlineAsmOptions::PRESERVES_FLAGS) {
+        generated_asm.push_str("    pushf\n");
+    }
```
Amanieu, line 89: Flags are not preserved by the calling convention so you don't need to save them here.

src/inline_asm.rs
```diff
@@ -0,0 +1,227 @@
...
+        return;
+    }
+
+    let mut slot_size = Size::from_bytes(0);
+    let mut clobbered_regs = Vec::new();
+    let mut inputs = Vec::new();
+    let mut outputs = Vec::new();
+
+    let mut new_slot = |reg_class| {
+        // FIXME align slot
```
Amanieu, line 29: You can assume that all register sizes are powers of 2 and that they are aligned to their size.

src/inline_asm.rs
```diff
@@ -0,0 +1,227 @@
...
+    writeln!(generated_asm, "{}:", asm_name).unwrap();
+
+    generated_asm.push_str(".intel_syntax noprefix\n");
+    generated_asm.push_str("    push rbp\n");
+    generated_asm.push_str("    mov rbp,rdi\n");
+    if !options.contains(InlineAsmOptions::PRESERVES_FLAGS) {
+        generated_asm.push_str("    pushf\n");
+    }
+
+    // Save clobbered registers
```
Amanieu, line 91: You should only save clobbered registers that are required to be saved by the calling convention.

src/inline_asm.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+fn expect_reg(reg_or_class: InlineAsmRegOrRegClass) -> InlineAsmReg {
+    match reg_or_class {
+        InlineAsmRegOrRegClass::Reg(reg) => reg,
+        InlineAsmRegOrRegClass::RegClass(class) => unimplemented!("{:?}", class),
+    }
+}
+
+fn reg_class_size(reg_class: InlineAsmRegClass) -> Size {
```
Amanieu, line 193: I think you can actually derive this from `reg_class.supported_types()`. Just take the largest size of all supported types for the register class.

src/inline_asm.rs
```diff
@@ -0,0 +1,227 @@
...
+    asm_name: &str,
+    arch: InlineAsmArch,
+    options: InlineAsmOptions,
+    template: &[InlineAsmTemplatePiece],
+    clobbered_regs: Vec<(InlineAsmReg, Size)>,
+    inputs: &[(InlineAsmReg, Size, Value)],
+    outputs: &[(InlineAsmReg, Size, CPlace<'_>)],
+) -> String {
+    let mut generated_asm = String::new();
+    writeln!(generated_asm, ".globl {}", asm_name).unwrap();
```
Amanieu, line 81: You want some additional metadata here, have a look at what rustc generates. In particular:
- You should create a section just for this function so the linker can discard it if it is not used. (`.section .text.foo,"ax",@progbits`). Use `.text` to switch back afterwards.
- Mark the symbol as a function type (`.type foo,@function`)
- Set the size of the symbol (`.size foo, foo_end-foo`)

src/inline_asm.rs
```diff
@@ -0,0 +1,227 @@
...
+                let reg = expect_reg(reg);
+                clobbered_regs.push((reg, new_slot(reg.reg_class())));
+                inputs.push((reg, new_slot(reg.reg_class()), crate::base::trans_operand(fx, in_value).load_scalar(fx)));
+                if let Some(out_place) = out_place {
+                    outputs.push((reg, new_slot(reg.reg_class()), crate::base::trans_place(fx, out_place)));
+                }
+            }
+            InlineAsmOperand::Const { value: _ } => todo!(),
+            InlineAsmOperand::SymFn { value: _ } => todo!(),
+            InlineAsmOperand::SymStatic { def_id: _ } => todo!(),
```
Amanieu, line 59: Is there a reason why you're not going through librustc_codegen_ssa's `FunctionCx::codegen_terminator`? `codegen_asm_terminator` does a bunch of necessary lowering for `Const`, `SymFn` and `SymStatic` that you will need to support them.

bjorn3: I don't yet use `rustc_codegen_ssa` for anything but the linking code. `fx` is of type `crate::common::FunctionCx`, not `rustc_codegen_ssa::mir::FunctionCx`.

src/inline_asm.rs
bjorn3: `reg_byte` registers have a size of 1. If you first have a `reg_byte` register and then a `reg` register then the `reg` register is misaligned.

Amanieu: If you assume that align == size then you can just round up the offset to a multiple of the size.

bjorn3: I misunderstood your comment. I thought you meant that the offset was already aligned in all cases. I have now implemented the align logic.

src/inline_asm.rs
bjorn3: Added a fixme.

src/inline_asm.rs
```diff
@@ -0,0 +1,219 @@
...
+
+        generated_asm.push_str("    pop rbp\n");
+        generated_asm.push_str("    ret\n");
+    } else {
+        generated_asm.push_str("    ud2\n");
+    }
+
+    generated_asm.push_str(".att_syntax\n");
+    writeln!(generated_asm, ".size {name}, .-{name}", name=asm_name).unwrap();
+    generated_asm.push_str(".popsection\n");
```
Amanieu, line 148: It's safer to just reset the section with `.text` here.

src/inline_asm.rs
```diff
@@ -0,0 +1,219 @@
...
+    _span: Span,
+    template: &[InlineAsmTemplatePiece],
+    operands: &[InlineAsmOperand<'tcx>],
+    options: InlineAsmOptions,
+) {
+    // FIXME add .eh_frame unwind info directives
+
+    if template.is_empty() {
+        // Black box
+        return;
```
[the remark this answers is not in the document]
bjorn3: I know. This is just to not error when using `black_box` on macOS and Windows. I haven't yet implemented assembly support for them. It doesn't matter that much right now as Cranelift barely optimizes things right now.

src/inline_asm.rs
```diff
@@ -0,0 +1,219 @@
...
+            .iter()
+            .map(|(ty, _)| ty.size())
+            .max()
+            .unwrap();
+        let align = rustc_target::abi::Align::from_bytes(reg_size.bytes()).unwrap();
+        slot_size = slot_size.align_to(align);
+        let offset = slot_size;
+        slot_size += reg_size;
+        offset
+    };
```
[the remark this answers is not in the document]
bjorn3: Added fixme

PR #1065, opened 2020-07-12
Add simple profile filter rust script

bjorn3:
This serves two functions:

1. Make it easier to interpret profiles of running cg_clif.
2. Demo the JIT backend of cg_clif.


bjorn3:
I recently read about the `#![...] /*` trick for making a file that is both valid rust and bash. I can't remember where though.

PR #1067, opened 2020-07-15
Use builtin ranlib instead of running external ranlib

### LINKED ISSUE #763: Reimplement ranlib in rust
Necessary for cross compiling, as using the native ranlib causes problems when cross compiling between systems using GNU and BSD style ar archives.

### LINKED ISSUE #869: Silence ranlib "the table of contents is empty" warning on macOS

bjorn3:
This makes it possible to create rlibs for any target without installing
a toolchain for the target

Fixes #763
Fixes <strike>#869</strike>


bjorn3:
For some reason builtin ranlib doesn't work correctly on macOS. I have disabled it on macOS for now.

PR #1079, opened 2020-08-28
Run cargo fmt on the whole project

CohenArthur:


src/abi/mod.rs
```diff
@@ -14,7 +14,10 @@ use crate::prelude::*;
 pub(crate) use self::returning::{can_return_to_ssa_var, codegen_return};
 
 // Copied from https://github.com/rust-lang/rust/blob/f52c72948aa1dd718cc1f168d21c91c584c0a662/src/librustc_middle/ty/layout.rs#L2301
-pub(crate) fn fn_sig_for_fn_abi<'tcx>(tcx: TyCtxt<'tcx>, instance: Instance<'tcx>) -> ty::PolyFnSig<'tcx> {
+pub(crate) fn fn_sig_for_fn_abi<'tcx>(
```
bjorn3, line 17: Please use `#[rustfmt::skip]` on this function. It is supposed to be a 1-to-1 copy of the rustc version. While I have to slightly change to to work with the different imports in this file, I would like to minimize the diff when I have to sync it with upstream again.

src/target_features_whitelist.rs
```diff
@@ -107,8 +107,10 @@ const POWERPC_ALLOWED_FEATURES: &[(&str, Option<Symbol>)] = &[
     ("vsx", Some(sym::powerpc_target_feature)),
 ];
 
-const MIPS_ALLOWED_FEATURES: &[(&str, Option<Symbol>)] =
```
bjorn3, line 110: Please add `#![rustfmt::skip]` at the top of this file. It is a literal copy a big chunk of the equivalent rustc file.

src/optimize/mod.rs
```diff
@@ -5,8 +5,7 @@ mod stack2reg;
 
 pub(crate) fn optimize_function<'tcx>(
     tcx: TyCtxt<'tcx>,
-    #[cfg_attr(not(debug_assertions), allow(unused_variables))]
-    instance: Instance<'tcx>,
+    #[cfg_attr(not(debug_assertions), allow(unused_variables))] instance: Instance<'tcx>,
```
bjorn3, line 8:
```suggestion
    instance: Instance<'tcx>,
```

It has been used in all cases for a while.

src/abi/mod.rs
CohenArthur: Fixed in 9c848e1

src/target_features_whitelist.rs
CohenArthur: Fixed in ed35bc8

src/target_features_whitelist.rs
```diff
@@ -1,3 +1,5 @@
+#[rustfmt::skip]
```
bjorn3, line 1:
```suggestion
#![rustfmt::skip]
```

CohenArthur: Oops, good catch

bjorn3 (approved):
Thanks!

bjorn3:
Forgot to merge.

PR #1083, opened 2020-09-23
Fix lld

bjorn3:
Lld doesn't allow dynamic relocations in readonly sections by default. This means that `.eh_frame` needs to use pcrel pointers. I also needed https://github.com/bytecodealliance/wasmtime/pull/2212 to fix readonly data objects with relocations that specify a custom section. As a bonus I also added support for https://github.com/bytecodealliance/wasmtime/pull/2218.

r? @philipc


bjorn3:
`lld` works now. `ld.bfd` doesn't work anymore.

bjorn3:
I needed to use `DW_EH_PE_sdata4` instead of `DW_EH_PE_udata4`.

bjorn3:
I will go ahead and merge this without waiting for a review.

Issue #1087, opened 2020-09-28
In-place binary/jit updating


lachlansneff:
Loris Cro and Andrew Kelley recently published a blog post (https://kristoff.it/blog/zig-new-relationship-llvm/) about Ziglang's new self-hosted compiler. Part of the post is about how they can do in-place patching of the binary to avoid doing linking when changing small amounts of code (or maybe larger amounts too, I'm not sure). This seems extremely powerful. Linking is consistently a significant bottleneck in compiling rust, and completely skipping it would probably result in crazy speedups for compiling.

I'm proposing that this project pursue this either in a similar way to Zig, by patching individual functions in the output binary, or by keeping the jit open and recompiling individual functions as the source changes.

bjorn3:
Thanks for pointing me to that post! Binary patching is probably hard to implement. Doing this for the JIT mode of cg_clif is much more possible. cranelift-simplejit currently doesn't support hot code swapping. I already wanted to implement this for lazy compilation in JIT mode though. Rustc doesn't have a `DepNode` that I can use to implement incremental compilation of single mono items, but that shouldn't be too hard to implement. https://github.com/rust-lang/rust/pull/76474 will make it possible to create a custom rustc driver that would listen for file changes and pass the necessary persistent state to the various runs of the codegen backend.

lachlansneff:
@bjorn3 That's fantastic, I'm glad to hear that the prerequisite components for this to work are mostly already on the roadmap.

As for the binary reloading, I don't think it'd actually be too bad, since the way that Zig does this is by including a function lookup table in the binary that can be modified very easily to redirect functions to new versions. It would require logic for each output format (mach, elf, pe, wasm, etc) though.

bjorn3:
I have been trying to revive an old branch for lazy compilation in jit mode. Currently the following program sometimes panics. (about one in five tries I think)

<details>

```rust
#![feature(
    no_core, start, lang_items, box_syntax, never_type, linkage,
    extern_types, thread_local
)]
#![no_core]
#![allow(dead_code, non_camel_case_types)]

extern crate mini_core;

use mini_core::*;
use mini_core::libc::*;

unsafe extern "C" fn my_puts(s: *const i8) {
    puts(s);
}

#[lang = "termination"]
trait Termination {
    fn report(self) -> i32;
}

impl Termination for () {
    fn report(self) -> i32 {
        unsafe {
            0
        }
    }
}

#[lang = "start"]
[...]

Issue #1066, opened 2020-07-13
Add experimental jit mode integration to cargo


bjorn3:
https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/jit.20support

m4b:
So would this mean:

```
CG_CLIF_JIT=1 cargo build
target/debug/foo
```
will:

1. produce a debug binary
2. the binary when run, will have instrumentation necessary to jit parts of the code in foo that would normally be AOT

bjorn3:
No, in the JIT mode the rustc invocation that "builds" the executable actually JITs it and then immediately runs it. No executable is written to the disk. I did imagine that the cargo command would be `cargo run --jit`.

m4b:
Oh interesting, is it possible to share the majority of the dep tree (e.g., i have 400 crates) when jitting, so a modify source, cargo jit loop would be faster?

And today, if i pass this flag to cargo cranelift and do cargo run, this won't do as you say above, right? It only currently works with explicit invocations of the cranelift using rustc, yes?

m4b:
Also, are there any plans to implement 2. I guess it's similar to adding something like a ~~java~~ Rust HotSpot in each binary ; )

bjorn3:
> Oh interesting, is it possible to share the majority of the dep tree (e.g., i have 400 crates) when jitting, so a modify source, cargo jit loop would be faster?

Currently only dylibs can be loaded by the JIT. I want to also make it work with rlibs though.

> And today, if i pass this flag to cargo cranelift and do cargo run, this won't do as you say above, right? It only currently works with explicit invocations of the cranelift using rustc, yes?

It will kinda work if you use your `cargo build` command, but build scripts will also be JITed, causing cargo to fail to run the executable it expects to be created.

> Also, are there any plans to implement 2. I guess it's similar to adding something like a java Rust HotSpot in each binary ; )

I may attempt to integrate the [Yorick](https://github.com/softdevteam/yk) meta tracer in the future. It does require explicit annotations about when to JIT what. It is also only meant to optimize language interpreters, not your average program.

PR #1089, opened 2020-09-29
Add custom rustc driver that uses cg_clif

bjorn3:
This makes the jit mode a bit more usable with cargo. This is also a prereq for the jit mode version of #1087.

Part of #1066


bjorn3:
This is by the way possible since https://github.com/rust-lang/rust/pull/76474.

PR #1094, opened 2020-10-16
Use architecture pointer size for AddConstructor

l4l:


src/backend.rs
```diff
@@ -132,10 +132,16 @@ impl AddConstructor for ObjectProduct {
         let init_array_section =
             self.object
                 .add_section(segment.to_vec(), b".init_array".to_vec(), SectionKind::Data);
+        let addres_size = self
```
bjorn3, line 135:
```suggestion
        let address_size = self
```
nit: typo

bjorn3 (approved):
Thanks!

src/backend.rs
```diff
@@ -132,10 +132,16 @@ impl AddConstructor for ObjectProduct {
...
+            .object
+            .architecture()
+            .address_size()
+            .expect("address_size must be known")
+            .bytes();
         self.object.append_section_data(
             init_array_section,
             &std::iter::repeat(0)
-                .take(8 /*FIXME pointer size*/)
+                .take(addres_size.into())
```
bjorn3, line 144:
```suggestion
                .take(address_size.into())
```

src/backend.rs
```diff
@@ -144,7 +150,7 @@ impl AddConstructor for ObjectProduct {
                 init_array_section,
                 object::write::Relocation {
                     offset: 0,
-                    size: 64, // FIXME pointer size
+                    size: addres_size * 8,
```
bjorn3, line 153:
```suggestion
                    size: address_size * 8,
```

src/backend.rs
l4l: Oops, thanks

PR #1096, opened 2020-10-27
Prefer numeric associated constants in example

bstrie:
Per their documentation, the `max_value()` and `min_value()` associated functions have been superseded by the `MAX` and `MIN` associated constants since Rust 1.43 and are considered "soft deprecated", with all uses currently being replaced in the rustc repo.


bjorn3:
Thanks! I didn't know it was soft deprecated.

PR #1108, opened 2020-11-04
Fix compiling with CARGO_TARGET_DIR

jyn514:
Previously it would error:

```
    Finished dev [unoptimized + debuginfo] target(s) in 26.93s
mv target/debug/main raytracer_cg_llvm
mv: cannot stat 'target/debug/main': No such file or directory
```


bjorn3:
Thanks!

Issue #1107, opened 2020-11-03
Run shellcheck on all scripts


bjorn3:

bjorn3:
Fixed in #1109

PR #1109, opened 2020-11-04
fix (many) shellcheck warnings

jakobhellermann:
Related issue: #1107 

The first commit fixes the 'machine applicable' shellcheck warnings, the second one fixes some manually.

There are still some warnings left:

```
In build_sysroot/prepare_sysroot_src.sh line 24:
for file in $(ls ../../patches/ | grep -v patcha); do
              ^-- SC2010: Don't use ls | grep. Use a glob or a for loop with a condition to allow non-alphanumeric filenames.
```
I wasn't sure why you were ignoring 'patcha' here. Is this still needed?

```
In scripts/config.sh line 31:
      RUN_WRAPPER='wine'
      ^---------^ SC2034: RUN_WRAPPER appears unused. Verify use (or export if used externally).
```
Was this supposed to be used somewhere?

```
In scripts/rustup.sh line 30:
	cg_clif=$(pwd)
        ^-----^ SC2034: cg_clif appears unused. Verify use (or export if used externally).
```
Same as above.

```
In scripts/tests.sh line 102:
    cat examples/regexdna-input.txt | ../build/cargo.sh run --example shootout-regex-dna | grep -v "Spawned thread" > res.txt
        ^-------------------------^ SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```
I think not using cat here makes readability worse.

### DIFF
--- a/scripts/cargo.sh
+++ b/scripts/cargo.sh
@@ -10,7 +10,7 @@ cmd=$1
 shift || true
 
 if [[ "$cmd" = "jit" ]]; then
-cargo "+${TOOLCHAIN}" rustc $@ -- --jit
+cargo "+${TOOLCHAIN}" rustc "$@" -- --jit
 else
-cargo "+${TOOLCHAIN}" "$cmd" $@
+cargo "+${TOOLCHAIN}" "$cmd" "$@"
 fi
--- a/scripts/cargo.sh
+++ b/scripts/cargo.sh
@@ -1,16 +1,16 @@
 #!/bin/bash
 
 dir=$(dirname "$0")
-source $dir/config.sh
+source "$dir/config.sh"
 
 # read nightly compiler from rust-toolchain file
-TOOLCHAIN=$(cat $dir/rust-toolchain)
+TOOLCHAIN=$(cat "$dir/rust-toolchain")
 
 cmd=$1
 shift || true
 
 if [[ "$cmd" = "jit" ]]; then
-cargo +${TOOLCHAIN} rustc $@ -- --jit
+cargo "+${TOOLCHAIN}" rustc $@ -- --jit
 else
-cargo +${TOOLCHAIN} $cmd $@
+cargo "+${TOOLCHAIN}" "$cmd" $@
 fi
--- a/scripts/config.sh
+++ b/scripts/config.sh
@@ -1,6 +1,6 @@
 set -e
 
-unamestr=`uname`
+unamestr=$(uname)
 if [[ "$unamestr" == 'Linux' ]]; then
    dylib_ext='so'
 elif [[ "$unamestr" == 'Darwin' ]]; then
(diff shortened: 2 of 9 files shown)


scripts/cargo.sh
```diff
@@ -10,7 +10,7 @@ cmd=$1
 shift || true
 
 if [[ "$cmd" = "jit" ]]; then
-cargo "+${TOOLCHAIN}" rustc $@ -- --jit
+cargo "+${TOOLCHAIN}" rustc "$@" -- --jit
```
bjorn3, line 13: The old behavior is deliberate. Separate arguments to `cargo.sh` should be passed as separate arguments to cargo.

scripts/config.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
bjorn3, line 1: This file should never be executed, only sourced. Please remove this line and revert permissions back.

test.sh
```diff
@@ -3,13 +3,13 @@ set -e
 
 export RUSTFLAGS="-Zrun_dsymutil=no"
 
-./build.sh --without-sysroot $@
+./build.sh --without-sysroot "$@"
```
bjorn3, line 6: Same re separate arguments.

scripts/config.sh
```diff
@@ -51,7 +53,8 @@ if [[ $(uname) == 'Darwin' ]]; then
    export RUSTFLAGS="$RUSTFLAGS -Clink-arg=-undefined -Clink-arg=dynamic_lookup"
 fi
 
-export LD_LIBRARY_PATH="$dir:$(rustc --print sysroot)/lib:$dir/target/out:$dir/sysroot/lib/rustlib/$TARGET_TRIPLE/lib"
+LD_LIBRARY_PATH="$dir:$(rustc --print sysroot)/lib:$dir/target/out:$dir/sysroot/lib/rustlib/$TARGET_TRIPLE/lib"
+export LD_LIBRARY_PATH
```
bjorn3, line 57: I slightly prefer the old version.

bjorn3 (changes requested):
Thanks!

> I wasn't sure why you were ignoring 'patcha' here. Is this still needed?

It allowed for quick disabling of specific patches by renaming them. I haven't used it in months, but it could be useful to keep.

> Was this supposed to be used somewhere?

`config.sh` is sourced by several files. `RUN_WRAPPER` is used by `tests.sh`.

> I think not using cat here makes readability worse.

I agree

scripts/cargo.sh
jakobhellermann: Actually I think this is correct: https://stackoverflow.com/a/3990540

"$@" expands to "$1" "$2" "$3".

The difference between `$@` and `"$@"` occurs when doing something like `./script "*" b c`, where the quoted variant will pass the * and the unquoted one will expand it.

scripts/config.sh
jakobhellermann: same as above

scripts/config.sh
jakobhellermann: I think the motivation was that if `rustc --print sysroot` fails, ` set -e` will not catch it inside an export, but I can change it.

scripts/config.sh
jakobhellermann: Done :+1:

scripts/cargo.sh
bjorn3: Didn't know that.

bjorn3 (approved):
Thanks!

PR #1111, opened 2020-11-12
use hardlinks instead of copying files

### LINKED ISSUE #1110: Use hardlinks instead of copying files
This will save disk space.

Copying currently happens at:

https://github.com/bjorn3/rustc_codegen_cranelift/blob/96c4542dc3c7001d5a28b05d067701f2173e9eb4/build.sh#L40-L42

and

https://github.com/bjorn3/rustc_codegen_cranelift/blob/96c4542dc3c7001d5a28b05d067701f2173e9eb4/build_sysroot/build_sysroot.sh#L39

jakobhellermann:
fixes #1110

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -37,9 +37,9 @@ fi
 rm -rf "$target_dir"
 mkdir "$target_dir"
 mkdir "$target_dir"/bin "$target_dir"/lib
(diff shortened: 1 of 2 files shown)


bjorn3:
Thanks!

PR #1120, opened 2020-12-25
Lazy compilation in jit mode

bjorn3:
Lazy compilation has the potential to significantly improve the startup time of a program. While functions have to be codegened when called, it is expected that a significant amount of all code is only required when an error occurs or only when the program is used in certain ways.

The basic approach is to first codegen a shim for each function. This shim calls the `__cg_clif_jit` function of cg_clif with a pointer to the `Instance` corresponding to the function for which it is a shim. `__cg_clif_jit` function then codegens this function and uses the hot code swapping support of SimpleJIT to redirect future calls to the function to the real version. Finally it calls the newly codegened function.

This is based on a suggestion by @flodiebold back in February. (https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/cranelift.20backend.20work/near/187645798) I made a quick prototype back then, but it was quite fragile and required too significant changes of the rest of cg_clif. The hot code swapping support of cranelift-jit I implemented in https://github.com/bytecodealliance/wasmtime/pull/2403 made it robust enough and reduced the required changes by a lot.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -68,7 +68,7 @@ $ $cg_clif_dir/build/cargo.sh jit
 or
 
 ```bash
-$ $cg_clif_dir/build/bin/cg_clif --jit my_crate.rs
+$ $cg_clif_dir/build/bin/cg_clif -Cllvm-args=mode=jit -Cprefer-dynamic my_crate.rs
 ```
 
 ### Shell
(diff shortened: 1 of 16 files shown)


bjorn3:
(to preserve the development history: d0c1fa2cbb742f30dc7b5daad3593acef5762629, 3cfe267b29e5e3b24d36f8643018873e3d54b445)

PR #1132, opened 2021-01-30
Don't set RUSTFLAGS when no specific linker is needed

Uriopass:
This allows the cargo.sh script to use the ~/.cargo/config if it exists it certain situations.


bjorn3:
Thanks!

bjorn3:
For future reference: https://github.com/bjorn3/rustc_codegen_cranelift/issues/1129#issuecomment-770202236

PR #1138, opened 2021-02-14
Test FreeBSD on CI

bjorn3:
Testing for one member of the BSD family should hopefully catch most issues with other members of the BSD family.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 # Settings
--- a/build.sh
+++ b/build.sh
@@ -34,13 +34,13 @@ done
 # Build cg_clif
 unset CARGO_TARGET_DIR
 unamestr=$(uname)
-if [[ "$unamestr" == 'Linux' ]]; then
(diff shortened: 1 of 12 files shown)


build.sh
```diff
@@ -34,13 +34,13 @@ done
 # Build cg_clif
 unset CARGO_TARGET_DIR
 unamestr=$(uname)
-if [[ "$unamestr" == 'Linux' ]]; then
+if [[ "$unamestr" == 'Linux' || "$unamestr" == "FreeBSD" ]]; then
```
bjorn3, line 37: ... this...

scripts/config.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 unamestr=$(uname)
-if [[ "$unamestr" == 'Linux' ]]; then
+if [[ "$unamestr" == 'Linux' || "$unamestr" == 'FreeBSD' ]]; then
```
bjorn3, line 6: ...and this.

bjorn3 (commented):
To add support for a new OS in the BSD family you should only need to change...

bjorn3:
While testing I noticed that Cirrus CI quite often restarts builds as GCE preempts the VM. It also takes several minutes before a build starts.

PR #1140, opened 2021-02-18
Remove support for the old x86 Cranelift backend

bjorn3:
This allows the new Cranelift atomic instructions to be used that lower directly to native atomic instructions, which makes it possible to remove the ugly hack based on a global lock for all "atomic" accesses.


bjorn3:
@Uriopass Could you try if the audio issue from https://github.com/bjorn3/rustc_codegen_cranelift/issues/1129 is fixed with this?

bjorn3:
I will leave this PR open for the next couple of days in case anyone has an objection to it.

Uriopass:
Sadly the audio issues are still present with this PR :/ I'll dive a bit more to find the root cause.
Native atomics are still good though :D

PR #1145, opened 2021-02-25
Support cross-compiling to Windows using MinGW

bjorn3:
Threads aren't supported yet as Cranelift doesn't support TLS for PE executables.

Part of https://github.com/bjorn3/rustc_codegen_cranelift/issues/977

Blocked on https://github.com/bytecodealliance/wasmtime/pull/2678

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        include:
+          - os: ubuntu-latest
+          - os: macos-latest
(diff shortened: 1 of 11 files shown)


kirawi:
Will MinGW be the only way to use it on Windows in the future, or is an MSVC compiled version planned?

bjorn3:
I assume that MSVC will also work, but the scripts for building cg_clif itself and the sysroot require a Unix system. You can manually build it on Windows or once https://github.com/rust-lang/rust/pull/81746 lands it will be available as rustup component on nightly.

bjorn3:
bytecodealliance/wasmtime#2678 has been merged.

PR #1163, opened 2021-04-23
Support -Ctarget-cpu

### LINKED ISSUE #690: Support -Ctarget-cpu

mominul:
Closes #690 
I am not familiar with the cranelift internals, thus I am not sure if this working properly, so does this need more work?

Thanks!

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -272,15 +272,28 @@ fn build_isa(sess: &Session, backend_config: &BackendConfig) -> Box<dyn isa::Tar
     let flags = settings::Flags::new(flags_builder);
 
     let variant = cranelift_codegen::isa::BackendVariant::MachInst;
-    let mut isa_builder = cranelift_codegen::isa::lookup_variant(target_triple, variant).unwrap();
+
+    let isa_builder = match sess.opts.cg.target_cpu.as_deref() {
+        Some("native") => {
+            let builder = cranelift_native::builder_with_options(variant, true).unwrap();
+            builder
+        }
+        Some(value) => {
+            let mut builder = cranelift_codegen::isa::lookup_variant(target_triple, variant).unwrap();
+            if let Err(_) = builder.enable(value) {
+                sess.fatal("The target cpu isn't currently supported by Cranelift.");
+            }
+            builder
+        }
+        None => {
+            let mut builder = cranelift_codegen::isa::lookup_variant(target_triple, variant).unwrap();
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -273,9 +273,14 @@ fn build_isa(sess: &Session, backend_config: &BackendConfig) -> Box<dyn isa::Tar
 
     let variant = cranelift_codegen::isa::BackendVariant::MachInst;
     let mut isa_builder = cranelift_codegen::isa::lookup_variant(target_triple, variant).unwrap();
-    // Don't use "haswell", as it implies `has_lzcnt`.macOS CI is still at Ivy Bridge EP, so `lzcnt`
-    // is interpreted as `bsr`.
-    isa_builder.enable("nehalem").unwrap();
+    
+    if let Some(target_cpu) = sess.opts.cg.target_cpu.as_ref() {
```
bjorn3, line 277: This will need to handle `-Ctarget-cpu=native`. I think you can use `cranelift_native::builder_with_options` for this as replacement of `cranelift_codegen::isa::lookup_variant` in case of `-Ctarget-cpu=native`.

Also can you use `sess.fatal()` when `.enable` returns an error telling the user that the target cpu isn't currently supported by Cranelift.

mominul:
I have pushed a commit that handles the `native` target-cpu variant. Can you review it again, please?

src/lib.rs
```diff
@@ -272,10 +272,28 @@ fn build_isa(sess: &Session, backend_config: &BackendConfig) -> Box<dyn isa::Tar
...
+
+    let isa_builder = match sess.opts.cg.target_cpu.as_deref() {
+        Some("native") => {
+            let builder = cranelift_native::builder_with_options(variant, true).unwrap();
+            builder
+        }
+        Some(value) => {
+            let mut builder = cranelift_codegen::isa::lookup_variant(target_triple, variant).unwrap();
+            if let Err(_) = builder.enable(value) {
+                sess.fatal("The target cpu isn't currently supported by Cranelift.");
```
bjorn3, line 284:
```suggestion
                sess.fatal("The specified target cpu isn't currently supported by Cranelift.");
```
nit

bjorn3 (approved):
Thanks! Just one small suggestion. Otherwise LGTM.

mominul:
I have updated the error message. Thanks to you also! :blush:

Issue #983, opened 2020-04-25
Register debuginfo to a debugger in jit mode.


bjorn3:
Wasmtime contains a lot of code to do this. It would be nice to extract this into a common crate.

bjorn3:
In 020bcb31dbb1e6a96575806a560bdc082bfa2bb6 the frame table information is now registered to the unwinder in jit mode.

bjorn3:
I asked for help on how to implement this on the gdb irc. It seems that gdb expects absolute virtual addresses for sections in the elf file. This is quite helpful as it allows the elf file containing the debuginfo to be created entirely without help from cranelift-jit. It may be necessary to copy the code into the elf file too though and point sh_offset to this copy. Thanks @simark for helping me figure this out!

https://gitlab.com/gnutools/binutils-gdb/-/blob/master/gdb/jit.c#L781-793

PR #1166, opened 2021-04-26
Multithreading support for lazy-jit

### LINKED ISSUE #1124: Support threads for lazy jit mode
It currently gives an ICE when needing to compile on a different thread. It should be possible to use a channel between the rustc thread and the thread requesting the compilation and then wait for the function to be compiled.

eggyal:
Fixes #1124, in which @bjorn3 commented:

> It should be possible to use a channel between the rustc thread and the thread requesting the compilation and then wait for the function to be compiled.

The problem I had using a channel was that I couldn't get the `Sender` to the child threads other than through using a global (in some way), which requires synchronisation... and then the channel just added additional synchronisation on top—which by that point was superfluous.  This PR uses a `Mutex<VecDeque>` for message passing instead.


bjorn3:
Thanks! I will take a look tomorrow.

eggyal:
Woke this morning with a realisation that it's probably better to use a channel after all—still requires synchronisation to get a clone, but that clone can then be stored thread-local which will then use far less locking.

eggyal:
Ran some tests with more threads and started hitting BAD_ACCESS errors... placing this in draft while I investigate.

eggyal:
Could it be that the memory allocated by Cranelift for the lazily-jitted function overlaps one or more pages in use by the running code such that when `finalize_definitions()` sets memory protection for the function, it's inadvertantly affecting unrelated allocations?

eggyal:
With a debugger attached, it looks like child threads hit `BAD_ACCESS` when the main rustc thread sets page permissions for some newly jitted function (at least, that's what the rustc thread is doing every single time I run it... whereas the child thread's `BAD_ACCESS` appears to move around: that said, #983 is definitely a hinderance to further diagnosis here!).

If that is the case, I guess that Cranelift and the children are sharing heap space and newly jitted functions sometimes end up on a page with data owned by child threads?  Then as soon as Cranelift sets the page permissions, some child thread hits `BAD_ACCESS` when attempting to modify its data?  If so then I guess, at least when lazily jitting, Cranelift should use an allocator that reserves entire pages for jitted code.

What I don't understand however is why lazy-jit mode doesn't appear to hit this when single-threaded?

eggyal:
This is what's failing (on MacOS):

```rust
fn main() {
    for n in 0..50 {
        std::thread::spawn(move || {
            println!("child {}", n);
        });
    }
}
```

I get output like this:

```
   Compiling lazy-jit-multithread-test v0.1.0 (/Users/eggyal/coding/lazy-jit-multithread-test)
Rustc codegen cranelift will JIT run the executable, because -Cllvm-args=mode=jit was passed
child 0
child 1
child 3
child 2
child 5
child 7
child 8
child 10
child 4
child 11
child 14
child 17
error: could not compile `lazy-jit-multithread-test`

Caused by:
  process didn't exit successfully: `/Users/eggyal/coding/rustc_codegen_cranelift/build/bin/cg_clif --crate-name lazy_jit_multithread_test --edition=2018 src/main.rs --error-format=json --json=diagnostic-rendered-ansi --crate-type bin --emit=dep-info,link -C embed-bitcode=no -C split-debuginfo=unpacked -C debuginfo=2 -Cllvm-args=mode=jit-lazy -Cprefer-dynamic -C metadata=e07189f45388568e --out-dir /Users/eggyal/coding/lazy-jit-multithread-test/target/debug/deps -C incremental=/Users/eggyal/coding/lazy-jit-multithread-test/target/debug/incremental -L dependency=/Users/eggyal/coding/lazy-jit-multithread-test/target/debug/deps -Clink-arg=-undefined -Clink-arg=dynamic_lookup -g -Csplit-debuginfo=packed -Clink-arg=-undefined -Clink-arg=dynamic_lookup` (signal: 10, SIGBUS: access to undefined memory)
```


Incomprehensibly, it appears to work okay if I name the threads:

```rust
fn main() {
    for n in 0..50 {
        std::thread::Builder::new()
            .name(format!("child cd{}", n))
            .spawn(move || {
                println!("child {}", n);
            })
            .unwrap();
    }
}
```

eggyal:
Have repeatedly run with a debugger attached, and the addresses being accessed that trigger this fault do not appear obviously corrupted... but they do all appear to be page-aligned.

eggyal:
After running a number of times with a debugger attached, it appears that the fault always arises upon a thread calling to the following assembly (which is always located at the start of a writable/non-executable page):

```asm
movq   %rsp, %rbp
movq   (%rdi), %rsi
testq  %rsi, %rsi
je     0x... ; jumps over call, to return
addq   $0x8, %rdi
movq   0x...(%rip), %rsi
callq  *%rsi
movq   %rbp, %rsp
popq   %rbp
retq
```

The address of this function was never returned by `__clif_jit_fn`, so I'm not too sure what it is or why the thread is trying to call it... alas #983 is hindering me here.

Investigation continues...

eggyal:
Ah.  I wonder.  When jitting a new function, does Cranelift place it on a page with an existing function thus necessitating that page's protection be temporarily changed from rx to rw?  Then if some other thread attempts to call a function on that page whilst it's in the temporary state, we'd get exactly this.  That also explains why the rustc thread is invariably in the process of setting page protection (during `finalize_definitions`) when these faults arise.

I guess the two possible solutions are:

1. only ever lazily jit functions onto a fresh page (could become quite expensive on memory);
2. effectively place the page behind some sort of `RwLock` and indirect every jitted function call via code to first obtain a read lock before entering the page.

I wonder how other multithreaded lazy jitters handle this...

eggyal:
> But it doesn't explain why naming the threads apparently makes a difference!!

It's not naming the thread that makes a difference, it's interpolating `n` into a string within the `format!` macro invocation that does; the same happens if one merely invokes `println!`.  My guess is that additional function calls move locations around such that the functions called by child threads no longer happen to fall on pages that are subsequently modified by Cranelift.

bjorn3:
I think https://github.com/bytecodealliance/wasmtime/pull/2786 is required for this to work. I need to implement https://github.com/bytecodealliance/wasmtime/pull/2786#issuecomment-815165530 though before it can be merged.

eggyal:
I removed my previous comments as I'm now encountering different variations... but yes, I'm sure you're right that that's the cause here.

eggyal:
I extended your branch to implement deferred GOT updates in https://github.com/eggyal/wasmtime/tree/atomic_jit — and that did indeed fix the problem here... but unless Cranelift is to itself run multithreaded, there's no chance of the GOT being updated from multiple threads so I don't think there's any need for `AtomicPtr`?  Certainly other threads are reading from the GOT and you don't want a dirty read, but those aren't atomic loads (unless it's your intention to introduce them?) and don't most (if not all?) architectures guarantee that reading `usize` will never be dirty anyway?

bjorn3:
Data races are UB in Rust and LLVM. LLVM is for example allowed to write arbitrary values (say 42) before writing the final value. This would crash the jitted code. Data races are not UB in Cranelift. For this reason the rust code needs to use atomic stores to prevent UB, while the Cranelift side doesn't need to use atomic loads to prevent UB. At least on all architectures currently supported by Cranelift.

By the way I think you will need to ensure that two simultaneous requests to compile the same function will result in only a single compilation.

bjorn3:
Instead of storing a `GotEntryKey` in `GotEntry` could you store an `*const AtomicPtr` pointing directly to the got entry instead? Also do you want me to incorporate the changes into my PR or do you want to open a replacement PR for Cranelift?

eggyal:
> Data races are UB in Rust and LLVM. LLVM is for example allowed to write arbitrary values (say 42) before writing the final value. This would crash the jitted code. Data races are not UB in Cranelift. For this reason the rust code needs to use atomic stores to prevent UB, while the Cranelift side doesn't need to use atomic loads to prevent UB. At least on all architectures currently supported by Cranelift.

Aha, got it.  Thanks.

> By the way I think you will need to ensure that two simultaneous requests to compile the same function will result in only a single compilation.

D'oh!  Well caught.  Considered holding some list of pending `*const Instance`s in a static, but since that would require locking on every jit request it brought me full circle back to the original idea of just handling messages via a static `Mutex<VecDeque>` and saving on the additional synchronisation provided by a channel.  Will push an update shortly.

> Instead of storing a GotEntryKey in GotEntry could you store an *const AtomicPtr pointing directly to the got entry instead? Also do you want me to incorporate the changes into my PR or do you want to open a replacement PR for Cranelift?

I've used `NonNull<AtomicPtr<u8>>`, which is quite neat as it just copies directly through the `Option`s, and of course enforces the additional semantics.  Have pushed that one to my branch already.  Don't mind however you want to do it... can open a PR to your branch in your repo if that's any help?

bjorn3:
I force pushed it directly to my PR branch.

eggyal:
> By the way I think you will need to ensure that two simultaneous requests to compile the same function will result in only a single compilation.

What do you think about the broad approach in 256a8929f2f6d50e074340fbd55f4b2e1ef08515?  The shim is definitely not entirely right as an assertion fails during lowering (hence I've marked this draft again), but before I dig any further into that I wanted to check if this approach is reasonable?

eggyal:
Requires exposure of `JITModule::get_got_address` in Cranelift, for which I'll submit a PR.

bjorn3:
https://github.com/bytecodealliance/wasmtime/pull/2786 has been merged. @eggyal can you please rebase this PR?

eggyal:
@bjorn3 Not sure whether you saw that I rebased this as requested, but pinging you just in case.

bjorn3:
Completely missed it. Thanks for the ping.

bjorn3:
Pushed a commit to test this feature.

bjorn3:
CI seems to be having some issues with installing wine for windows testing. Merging as I don't expect any failures from this CI job.

PR #1169, opened 2021-04-30
Set correct segment from #[link_section] for MachO

### LINKED ISSUE #1168: #[link_section]: set correct segment for Mach-O files
There's a FIXME comment in the source code

https://github.com/bjorn3/rustc_codegen_cranelift/blob/88901ca9d438787cf68874c9c2a65d0a95bbc010/src/constant.rs#L376-L379

but I couldn't find an issue to track it.

For information: I hit this when building the log crate with dev channel rustc.  sval-rs/value-bag#22 resulted in use of the ctor crate, which uses `#[link_section = "__DATA,__mod_init_func"]` on MacOS: because cg_clif placed the whole string in the section name, gimli/object subsequently panicked while emitting MachO.

Is it not sufficient to split the link_section on `,` for all targets?

eggyal:
Closes #1168

### DIFF
--- a/src/constant.rs
+++ b/src/constant.rs
@@ -374,8 +374,19 @@ fn define_all_allocs(tcx: TyCtxt<'_>, module: &mut dyn Module, cx: &mut Constant
         data_ctx.set_align(alloc.align.bytes());
 
         if let Some(section_name) = section_name {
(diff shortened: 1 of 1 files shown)


bjorn3 (approved):
Sorry, forgot about this.

PR #1178, opened 2021-06-14
Implement llvm.x86.addcarry.64 and llvm.x86.subborrow.64

shamatar:
My first attempt to contribute to the compiler.

It works (builds `num-bigint` version 0.4 and passes some long additions/subtractions test), my main concern is whether I should optimize manually for a constant value of carry provided (constant `false` case mainly for starts of addition chains) or is it done by other codegen functions internally

### DIFF
--- a/src/intrinsics/llvm.rs
+++ b/src/intrinsics/llvm.rs
@@ -106,6 +106,26 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
             let dest = CPlace::for_ptr(Pointer::new(mem_addr), a.layout());
             dest.write_cvalue(fx, a);
         };
+        "llvm.x86.addcarry.64", (c c_in, c a, c b) {
+            llvm_add_sub_with_carry(
+                fx,
+                "llvm.x86.addcarry.64",
+                ret,
+                c_in,
+                a,
+                b
+            );
+        };
+        "llvm.x86.subborrow.64", (c b_in, c a, c b) {
+            llvm_add_sub_with_carry(
+                fx,
+                "llvm.x86.subborrow.64",
+                ret,
+                b_in,
+                a,
+                b
+            );
+        };
     }
 
     if let Some((_, dest)) = destination {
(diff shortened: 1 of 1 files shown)


bjorn3:
Thanks for the PR! I will take a look tomorrow.

src/intrinsics/llvm.rs
```diff
@@ -121,3 +141,60 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
...
+        fx,
+        bin_op,
+        c,
+        cb_in_as_u64,
+    );
+    let c = int1.value_field(fx, mir::Field::new(0));
+    let cb1 = int1.value_field(fx, mir::Field::new(1));
+    
+    // carry0 | carry1 -> carry or borrow respectively
+    let cb = crate::num::codegen_bool_binop(
```
bjorn3, line 189: Can you call `load_scalar` on `cb0` and `cb1` first and then use `fx.bcx.ins().bor(cb0, cb1)`?

src/intrinsics/llvm.rs
```diff
@@ -121,3 +141,60 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
...
+        BinOp::BitOr,
+        cb0,
+        cb1,
+    );
+    let layout = fx.layout_of(fx.tcx.mk_tup([fx.tcx.types.u8, fx.tcx.types.u64].iter()));
+    let cb_out = cb.load_scalar(fx);
+    let c = c.load_scalar(fx);
+    let val = CValue::by_val_pair(cb_out, c, layout);
+    ret.write_cvalue(fx, val);
+}
```
bjorn3, line 200: nit: missing trailing newline

src/intrinsics/llvm.rs
```diff
@@ -121,3 +141,60 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
...
+    // carry0 | carry1 -> carry or borrow respectively
+    let cb = crate::num::codegen_bool_binop(
+        fx,
+        BinOp::BitOr,
+        cb0,
+        cb1,
+    );
+    let layout = fx.layout_of(fx.tcx.mk_tup([fx.tcx.types.u8, fx.tcx.types.u64].iter()));
+    let cb_out = cb.load_scalar(fx);
+    let c = c.load_scalar(fx);
```
bjorn3, line 197: Can you immediately call `load_scalar` before assigning to `c` above?

src/intrinsics/llvm.rs
```diff
@@ -106,6 +106,26 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
             let dest = CPlace::for_ptr(Pointer::new(mem_addr), a.layout());
             dest.write_cvalue(fx, a);
         };
+        "llvm.x86.addcarry.64", (c c_in, c a, c b) {
```
bjorn3, line 109: If you use `v c_in` you will get a `Value` instead of `CValue<'tcx>` to pass in to `llvm_add_sub_with_carry`.

src/intrinsics/llvm.rs
```diff
@@ -106,6 +106,26 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
...
+            llvm_add_sub_with_carry(
+                fx,
+                "llvm.x86.addcarry.64",
```
bjorn3, line 112: Please pass in `BinOp::Add` instead.

src/intrinsics/llvm.rs
shamatar: In this case it's not possible to add a check about the type and layout of the carry/borrow. If this check is not too important then I can take a "Value" instead

shamatar:
Separately I should note that there is another missing instrinsic `llvm.x86.xgetbv` when building the libstd and co, somewhere in `objekt`

bjorn3:
xgetbv is not something that can be emulated. It reads cpu flags. It is only used by is_x86_feature_detected if cpuid says that cpu_osxsave is supported, which the cpuid instruction emulated by cg_clif doesn't say.

src/intrinsics/llvm.rs
bjorn3: It isn't too important to check the type. LLVM intrinsics can't be directly called without a nightly feature and any mistake in the type would likely cause LLVM to crash if using cg_llvm.

shamatar: Ok, I'll remove it then

src/intrinsics/llvm.rs
```diff
@@ -121,3 +141,45 @@ pub(crate) fn codegen_llvm_intrinsic_call<'tcx>(
 // llvm.x86.avx2.pshuf.b
 // llvm.x86.avx2.psrli.w
 // llvm.x86.sse2.psrli.w
+// llvm.x86.addcarry.64
+// llvm.x86.subborrow.64
```
bjorn3, line 145:
```suggestion
```
This is a list of intrinsics I have seen in the wild, but haven't implemented yet.

shamatar:
I think `blake2_simd` uses:
 llvm.x86.avx.storeu.dq.256
llvm.x86.sse41.pblendw
llvm.x86.ssse3.pshuf.b.128
llvm.x86.avx2.pshuf.b
llvm.x86.avx2.pslli.d
llvm.x86.avx2.psrli.d
llvm.x86.avx2.vperm2i128

bjorn3 (approved):

bjorn3:
Thanks!

Issue #1176, opened 2021-06-05
Rewrite build system in rust


bjorn3:
This would make it easier to build on windows.

The build system should keep allowing changes to be made to the list of things to do while a build is running by re-reading a config file at various points in the build process. The current shell scripts can be edited while they are running, which saves me a lot of time when I realize mid-build that a certain step can or can't be skipped.

# Todo

* [x] Rewrite test.sh in rust
* [ ] ~~Rewrite clean_all.sh in rust~~ (a program can't remove itself on windows)
* [ ] Add a windows variant of clean_all.sh
* [ ] Rewrite other shell scripts in rust (low priority)

afonso360:
I'm planning on moving test.sh over to y.rs, since its really hard to run tests on windows right now.

bjorn3:
I would like if it remains possible to disable tests while it is running. I often use the fact that bash reads shell script commands just before executing them to disable tests that I don't need while it is running in order to save time. (I often forget to do this before I started the tests.) Aborting and restarting would require rerunning tests again, which wastes time.

One way to do this I think would be to have a dsl for specifying tests and have the test runner parse the test specification after every step, skipping tests that have already succeeded.

afonso360:
What do you think about using the existing `config.txt` system?

We can add a key for each testsuite such as:
```
test.mod_bench=true
test.simple-raytracer=true
test.libcore=false
```

The infrastructure is already there, we just need to query each key before running the test.

bjorn3:
That would also work and is probably simpler to implement.

bjorn3:
The main build system is now entirely written in rust. For `clean_all.sh` it is easy to manually run the commands. And for the remaining shell scripts in `scripts/` I don't think there is much of a benefit to rewriting them as they are either mostly used on CI or just by myself.

PR #1180, opened 2021-06-19
Rewrite part of the build system in rust

bjorn3:
This allows building cg_clif on Windows for the x86_64-pc-windows-gnu target. The MSVC toolchain is not yet supported by cg_clif itself. ~~Note that `cargo.sh` is not yet ported, so you will have to manually set the `RUSTC` env var to point to cg_clif.exe in the `build/bin` dir. Leaving as draft until `cargo.sh` is ported.~~

# Todo (future PR)

* [x] Rewrite `cargo.sh` in rust (high priority)
* [ ] Rewrite `test.sh` in rust
* [ ] Rewrite `clean_all.sh` in rust
* [ ] Rewrite other shell scripts in rust (low priority)

cc #1176


bjorn3:
I rewrote cargo.sh in rust.

PR #1183, opened 2021-07-02
AArch64 support on Linux

bjorn3:
Thanks to @afonso360 Cranelift now supports everything on AArch64 that cg_clif needs when compiling for Linux. CI is currently testing cross-compilation. Native may work too, though I don't know if that would hit https://github.com/bjorn3/rustc_codegen_cranelift/issues/1060.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,6 +55,9 @@ jobs:
         git config --global user.name "User"
         ./y.rs prepare
 
+    - name: Build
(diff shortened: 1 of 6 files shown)


afonso360:
🎉  I'm happy to see this! I have ran into some issues running `test_rustc_tests.sh` natively, but I haven't had time to properly investigate. At a first glance it didn't seem to be anything aarch64 specific though.

I'll look into it again

PR #1191, opened 2021-08-04
change instructions for manually compiling y.rs

gheoan:
This prevents an error on windows where the `build_sysroot` function was trying to delete `y.exe`.


bjorn3:
Thanks!

PR #1227, opened 2022-05-09
Only enable JIT tests on x86_64

Kmeakin:
Cranelift currently only supports JIT on x86_64 targets.
Disable JIT tests on all other targets, so that failing tests are
ignored.

Originally reported in https://github.com/rust-lang/rust/pull/96437.


bjorn3:
Thanks!

PR #1252, opened 2022-07-30
Move test script to y.rs

afonso360:
👋  Hey,

This PR moves the tests scripts over to a subcommand on `y.rs`.

We now use `config.txt` to check which tests we should run. This allows the user to edit the file while the testsuite is running to skip running some tests.

Another change is that instead of passing `no_sysroot`/`base_sysroot`/`extended_sysroot` as a test argument, we also check those as keys in `config.txt`. Let me know if this is a feature that we want to keep as it was before @bjorn3.

cc: #1176


build_system/mod.rs
```diff
@@ -115,14 +122,27 @@ pub fn main() {
         process::exit(1);
     }
 
-    let cg_clif_build_dir =
+    let cg_clif_build_dir = Path::new("target").join(&host_triple).join(&channel);
+
+    if command == Command::Test {
+        // TODO: Should we also build_backend here?
```
bjorn3, line 128: Yes. It should build the backend first. Could you turn this `if` into a match and move build_backend before it?

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+
+                let input = fs::read_to_string(PathBuf::from("examples/regexdna-input.txt")).unwrap();
+                let expected_path = PathBuf::from("examples/regexdna-output.txt");
+                let expected = fs::read_to_string(&expected_path).unwrap();
+
+                let output = spawn_and_wait_with_input(run_cmd, input);
+                // Make sure `[codegen mono items] start` doesn't poison the diff
+                let output = output.lines()
+                    .filter(|line| !line.contains("codegen mono items"))
+                    .filter(|line| !line.contains("Spawned thread"))
```
bjorn3, line 208: Not sure filtering "Spawned thread" is still necessary. I think not.

build_system/build_sysroot.rs
```diff
@@ -35,11 +37,17 @@ pub(crate) fn build_sysroot(
 
     // Build and copy rustc and cargo wrappers
     for wrapper in ["rustc-clif", "cargo-clif"] {
+        let wrapper_name = if cfg!(windows) {
```
bjorn3, line 40: This should check the target. There is basic support for cross-compiling in the build system. Or better it should use `get_file_name` too. By the way `get_file_name` needs a target argument I think.

build_system/build_backend.rs
```diff
@@ -38,6 +37,4 @@ pub(crate) fn build_backend(
 
     eprintln!("[BUILD] rustc_codegen_cranelift");
     super::utils::spawn_and_wait(cmd);
-
-    Path::new("target").join(host_triple).join(channel)
```
bjorn3, line 42: Could you revert this change?

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+
+impl TestCase {
+    const fn new(config: &'static str, func: &'static dyn Fn(&TestRunner)) -> Self {
+        Self { config, func }
+    }
+}
+
+const NO_SYSROOT_SUITE: &[TestCase] = &[
+    TestCase::new("build.mini_core", &|runner| {
+        runner.run_rustc(["example/mini_core.rs", "--crate-name", "mini_core", "--crate-type", "lib,dylib", "--target", &runner.target_triple]);
```
bjorn3, line 26: These now all miss the `[BUILD] mini_core` output lines, right?

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+        eprintln!("[SKIP] extended_sysroot tests");
+    }
+
+    Ok(())
+}
+
+
+fn remove_out_dir() -> Result<()> {
+    let out_dir = Path::new("target").join("out");
+    Ok(fs::remove_dir_all(out_dir)?)
```
bjorn3, line 321: Could you inline this function?

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+
+        let mut out_dir = root_dir.clone();
+        out_dir.push("target");
+        out_dir.push("out");
+
+        let is_native = host_triple == target_triple;
+        let jit_supported = target_triple.contains("x86_64") && is_native;
+
+        let env_rust_flags = env::var("RUSTFLAGS").ok();
+        let env_run_wrapper = env::var("RUN_WRAPPER").ok();
```
bjorn3, line 346: `RUN_WRAPPER` is not a public env var. It is always overwritten and it probably shouldn't even have been exported as env var. Could you avoid checking it?

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+fn remove_out_dir() -> Result<()> {
+    let out_dir = Path::new("target").join("out");
+    Ok(fs::remove_dir_all(out_dir)?)
+}
+
+struct TestRunner {
+    root_dir: PathBuf,
+    out_dir: PathBuf,
+    jit_supported: bool,
+    rust_flags: Vec<String>,
```
bjorn3, line 328: I think `rust_flags` should be represented as `String`, as this is how it will be passed to cargo anyway.

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+            rust_flags: rust_flags.iter().map(|s| s.to_string()).collect(),
+            run_wrapper: run_wrapper.iter().map(|s| s.to_string()).collect(),
+            host_triple,
+            target_triple,
+        }
+    }
+
+    pub fn run_testsuite(&self, tests: &[TestCase]) {
+        for &TestCase { config, func } in tests {
+            let is_jit_test = config.contains("jit");
```
bjorn3, line 387: I think this should check that the tag is an exact match to `"jit"`.

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        let mut rustc_clif = self.root_dir.clone();
+        rustc_clif.push("build");
+        rustc_clif.push("rustc-clif");
+        if cfg!(windows) {
+            rustc_clif.set_extension("exe");
+        }
```
bjorn3, line 428: This should use `get_file_name`.

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        let mut cargo_clif = self.root_dir.clone();
+        cargo_clif.push("build");
+        cargo_clif.push("cargo-clif");
+        if cfg!(windows) {
+            cargo_clif.set_extension("exe");
+        }
```
bjorn3, line 490: Same

build_system/tests.rs
```diff
@@ -0,0 +1,504 @@
...
+                // Make sure `[codegen mono items] start` doesn't poison the diff
+                let output = output.lines()
+                    .filter(|line| !line.contains("codegen mono items"))
+                    .filter(|line| !line.contains("Spawned thread"))
+                    .chain(Some("")) // This just adds the trailing newline
+                    .collect::<Vec<&str>>()
+                    .join("\r\n");
+
+
+                if output != expected {
```
bjorn3, line 214: Somewhere this is going wrong on CI:

```
--- res.txt	2022-07-30 10:22:41.404652369 +0000
+++ examples/regexdna-output.txt	2019-07-21 03:01:44.000000000 +0000
@@ -1,13 +1,13 @@
-agggtaaa|tttaccct 0
-[cgt]gggtaaa|tttaccc[acg] 3
-a[act]ggtaaa|tttacc[agt]t 9
-ag[act]gtaaa|tttac[agt]ct 8
-agg[act]taaa|ttta[agt]cct 10
-aggg[acg]aaa|ttt[cgt]ccct 3
-agggt[cgt]aa|tt[acg]accct 4
-agggta[cgt]a|t[acg]taccct 3
-agggtaa[cgt]|[acg]ttaccct 5
-
-101745
-100000
-133640
+agggtaaa|tttaccct 0
+[cgt]gggtaaa|tttaccc[acg] 3
+a[act]ggtaaa|tttacc[agt]t 9
+ag[act]gtaaa|tttac[agt]ct 8
+agg[act]taaa|ttta[agt]cct 10
+aggg[acg]aaa|ttt[cgt]ccct 3
+agggt[cgt]aa|tt[acg]accct 4
+agggta[cgt]a|t[acg]taccct 3
+agggtaa[cgt]|[acg]ttaccct 5
+
+101745
+100000
+133640
```

.github/workflows/main.yml
```diff
@@ -103,7 +103,7 @@ jobs:
         # Enable extra checks
         export CG_CLIF_ENABLE_VERIFIER=1
 
-        ./test.sh
+        ./y.rs test
```
bjorn3, line 106: I think the job running on windows could now run tests too.

bjorn3 (changes requested):
Thanks a lot!

> Another change is that instead of passing no_sysroot/base_sysroot/extended_sysroot as a test argument, we also check those as keys in config.txt. Let me know if this is a feature that we want to keep as it was before @bjorn3.

I don't recall every having invoked scripts/tests.sh directly.

build_system/tests.rs
afonso360: I didn't see any of those in the output, ill remove this.

build_system/tests.rs
afonso360: We emit those lines from the `TestCase` name. The tag gets upper cased, and the rest of the line is the test name.

It gives a similar output, although not exactly equal.

build_system/rustc_info.rs
```diff
@@ -43,14 +43,16 @@ pub(crate) fn get_default_sysroot() -> PathBuf {
     Path::new(String::from_utf8(default_sysroot).unwrap().trim()).to_owned()
 }
 
-pub(crate) fn get_file_name(crate_name: &str, crate_type: &str) -> String {
+pub(crate) fn get_file_name(crate_name: &str, crate_type: &str, target: &str) -> String {
```
afonso360, line 46: Turns out we always need the host triple when calling this. Would you like me to revert this change?

build_system/tests.rs
```diff
@@ -0,0 +1,491 @@
...
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        let mut rustc_clif = self.root_dir.clone();
+        rustc_clif.push("build");
+        rustc_clif.push(get_wrapper_file_name("rustc-clif", "bin", &self.host_triple));
+
+        let mut cmd = Command::new(rustc_clif);
+        cmd.args(self.rust_flags.split_whitespace());
```
afonso360, line 421: I really don't know what's going on here. If we don't pass the flags manually as args they never get processed. Despite us explicitly passing it in the env below.

Any ideas?

afonso360:
Sorry for the slightly messy PR and thanks for reviewing! I'm still learning my way around cg_clif.

It looks like the windows build is broken, is there a way to mark it as non blocking for CI?

bjorn3:
I think reverting back to only running builds is best until all tests are fixed. Otherwise no windows artifacts will be released anymore.

build_system/rustc_info.rs
bjorn3: It would be necessary when cross compiling the codegen backend and not just the sysroot, but given that cross compiling doesn't work due to rustc using proc macros (which are not compiled for the build triple), I think it is fine to revert this back.

build_system/tests.rs
bjorn3: RUSTFLAGS is only recognized by cargo and not by rustc.

bjorn3:
Thanks a lot! I pushed a commit to rustfmt the build system.

PR #1253, opened 2022-08-01
Fix `no_sysroot` testsuite for MSVC environments

afonso360:
👋 Hey,

This PR fixes the `no_sysroot` testsuite for windows.

It disables JIT tests. I haven't investigated those, but we seem to have some linking issues.

Links with the correct library for `printf` in `ucrt`.

Uses the MSVC `CreateThread` functions for the TLS tests.

### DIFF
--- a/example/mini_core_hello_world.rs
+++ b/example/mini_core_hello_world.rs
@@ -139,7 +139,7 @@ pub struct bool_11 {
     field10: bool,
 }
 
-extern "C" fn bool_struct_in_11(arg0: bool_11) {}
+extern "C" fn bool_struct_in_11(_arg0: bool_11) {}
 
 #[allow(unreachable_code)] // FIXME false positive
 fn main() {
@@ -375,6 +375,7 @@ struct pthread_attr_t {
 }
 
 #[link(name = "pthread")]
+#[cfg(not(target_env="msvc"))]
 extern "C" {
     fn pthread_attr_init(attr: *mut pthread_attr_t) -> c_int;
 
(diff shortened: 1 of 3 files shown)


example/mini_core_hello_world.rs
```diff
@@ -375,6 +375,7 @@ struct pthread_attr_t {
 }
 
 #[link(name = "pthread")]
+#[cfg(not(target_env="msvc"))]
```
bjorn3, line 378: Maybe `#[cfg(unix)]` instead?

example/mini_core_hello_world.rs
```diff
@@ -391,6 +392,86 @@ extern "C" {
...
+
+#[link(name = "msvcrt")]
+#[cfg(target_env="msvc")]
+extern "C" {
+    fn WaitForSingleObject(
+        hHandle: LPVOID,
+        dwMilliseconds: DWORD
+    ) -> DWORD;
+
+    fn  CreateThread(
```
bjorn3, line 409:
```suggestion
    fn CreateThread(
```

example/mini_core_hello_world.rs
```diff
@@ -391,6 +392,86 @@ extern "C" {
...
 }
 
+type DWORD = u32;
+type LPDWORD = *mut u32;
+
+type LPVOID = *mut c_void;
+type HANDLE = *mut c_void;
+
+#[link(name = "msvcrt")]
+#[cfg(target_env="msvc")]
```
bjorn3, line 402: Do these work with MinGW too? If so could you make this `#[cfg(windows)]`?

example/mini_core_hello_world.rs
```diff
@@ -391,6 +392,86 @@ extern "C" {
...
+        lpParameter: LPVOID,
+        dwCreationFlags: DWORD,
+        lpThreadId: LPDWORD
+    ) -> HANDLE;
+}
+
+enum Thread {
+    Windows(HANDLE),
+    Pthread(pthread_t)
+}
```
bjorn3, line 422: I think a struct with `#[cfg]` fields would be nicer.

example/mini_core_hello_world.rs
```diff
@@ -391,6 +392,86 @@ extern "C" {
...
+    unsafe fn join(self) {
+        match self {
+            #[cfg(not(target_env="msvc"))]
+            Thread::Pthread(thread) => {
+                let mut res = 0 as *mut c_void;
+                pthread_join(thread, &mut res);
+            }
+            #[cfg(target_env="msvc")]
+            Thread::Windows(handle) => {
+                let wait_time = 5000; // in milliseconds
```
bjorn3, line 464: Is it possible to have no timeout?

afonso360: Sort of. 0 makes this an async operation. Otherwise the value usually used is the `INFINITE` macro which is essentially just `0xffffffff`. I'll change it to that.

PR #1254, opened 2022-08-03
Use native cranelift instructions when lowering float intrinsics

afonso360:
👋 Hey,

This is a fix for the `LNK2001: unresolved external symbol fabsf` issue with the `base_sysroot` testsuite on MSVC.

Instead of always emitting a libcall we use cranelift's native instructions which don't emit a libcall (In processors with SSE4.1, although we can fix this with a different lowering if its important). This should also provide a nice performance boost for these operations.

msvcrt does not provide `fabsf` so we avoid linking to it here. An alternative solution could be to provide it ourselves, but this felt like a better solution.

We now pass the entire `base_sysroot` testsuite on MSVC 🎉

### DIFF
--- a/src/intrinsics/mod.rs
+++ b/src/intrinsics/mod.rs
@@ -301,7 +301,38 @@ fn codegen_float_intrinsic_call<'tcx>(
         _ => unreachable!(),
     };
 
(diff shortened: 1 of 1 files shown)


bjorn3:
Thanks!

PR #1266, opened 2022-08-18
Refactorings for enabling parallel compilation (part 2)

bjorn3:
This PR contains a bunch of refactorings to remove dependencies on TyCtxt and Session for various functions that will in the future be called on background threads as those types are only available on the main thread. Most of this PR concerns the debuginfo generation code, but there are also a couple of refactorings to other code.

Part of https://github.com/bjorn3/rustc_codegen_cranelift/issues/652

See https://github.com/bjorn3/rustc_codegen_cranelift/pull/1264 for the previous PR in this series.

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -10,6 +10,7 @@ use rustc_middle::ty::SymbolName;
 use indexmap::IndexSet;
 
 use crate::constant::ConstantCx;
+use crate::debuginfo::FunctionDebugContext;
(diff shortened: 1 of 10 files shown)


bjorn3:
Before merging this I'm going to benchmark it first as I'm not entirely sure about the perf implications of 6421427.

bjorn3:
Doesn't seem to have changed much.

PR #1267, opened 2022-08-21
Use native scalar `fma` instruction

afonso360:
Cranelift 0.87 now supports [lowering `fma` as a libcall on x86](https://github.com/bytecodealliance/wasmtime/commit/709716bb8e6adaf7e65f3497168af23ce0cf09ef).
With 0.88 enabling the native x86 instruction under the `has_fma` flag.

aarch64 and s390x already support this as a native instruction, so it's nice that we emit it for those.

We can't lower the SIMD version using the `fma` instruction since the lowering can fail if the x86 `has_fma` flag is not enabled. Cranelift doesn't yet know how to fallback for these cases.

We need to wait for the 0.87 release before merging this.

### DIFF
--- a/src/intrinsics/mod.rs
+++ b/src/intrinsics/mod.rs
@@ -303,6 +303,12 @@ fn codegen_float_intrinsic_call<'tcx>(
 
     let layout = fx.layout_of(ty);
     let res = match intrinsic {
+        sym::fmaf32 | sym::fmaf64 => {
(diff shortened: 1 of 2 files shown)


bjorn3:
Updated to Cranelift 0.87.0 in https://github.com/bjorn3/rustc_codegen_cranelift/commit/b14c7337db5861288afc3b26ba006859ed052d39. Can you please rebase?

PR #1269, opened 2022-08-21
JIT: Use exports to get symbol names for PE file formats

afonso360:
It looks like we should use the exports table to import symbols for PE file format files.

The JIT now imports the correct symbols for `jit.mini_core_hello_world` but still crashes while running it with a `Access violation - code c0000005`. Its a bit harder to understand where or what exactly the issue is since the debugger does not pick up any debug info.

### DIFF
--- a/src/driver/jit.rs
+++ b/src/driver/jit.rs
@@ -318,13 +318,32 @@ fn load_imported_symbols_for_jit(
 
     let mut imported_symbols = Vec::new();
     for path in dylib_paths {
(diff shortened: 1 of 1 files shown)


philipc:
The intent of the `object` crate is that `exports` should work for ELF and Mach-O as well. If it doesn't then that would be worth a bug report.

bjorn3:
Didn't know about the exports method. It seems that it can't distinguish between code, data and tls exports though.Tls exports have to be ignored as they can't be dlsym'd.

PR #1270, opened 2022-08-24
prevent unexpected formatting by rustfmt

csmoe:
While open this file in an editor which is configured with things like autoformat, lines between `set -e` and `*/` will be formatted into typical rust comments.


csmoe:
It's okay with `normalize_comments = false`.

Issue #977, opened 2020-04-22
Windows support


FireFlyForLife:
I believe cranelift is cross-platform. At least on windows.

However I only found Bash scripts for building with cargo for example. Is this a limitation, or am I missing something?

If a port from Bash to Batch (or something else like python) is needed I am of course happy to help!

bjorn3:
I have been focusing on Linux and macOS. There are several places where Windows support is missing:

* [`Abi::System`](https://github.com/bjorn3/rustc_codegen_cranelift/blob/5a6d9e1f1b2d8bca214c72d71f54a45141587106/src/abi/mod.rs#L89): This is equivalent to `Abi::C` on x86_64 I think, but I am not sure.
* [`.rlib`](https://github.com/bjorn3/rustc_codegen_cranelift/blob/5a6d9e1f1b2d8bca214c72d71f54a45141587106/src/archive.rs): The `rust-ar` crate only supports BSD and Gnu style `ar` archives. I believe Windows uses a slightly different format.
* The bash scripts, as you already noticed. (only `config.sh`, `prepare.sh`, `test.sh` and `cargo.sh` need to be ported. `rustup.sh` is just for me to switch to a new rustc version more easily. I haven't used `flamegraph.sh` in ages.)
* Probably other things.

I did some experimentation on the [`windows_support`](https://github.com/bjorn3/rustc_codegen_cranelift/tree/windows_support) branch, but it currently fails as I used `bash` to run the scripts and `bash` turns `C:\` into `/c/`, which rustc doesn't recognize. As I don't use Windows, I have to wait for CI every time I try to fix something.

I don't think I will add Windows support soon, but PR's are of course welcome.

bjorn3:
I got working support for cross compilation from Linux to MinGW on the `wip_windows_support` branch. For reference: see the comment thread starting at https://github.com/bytecodealliance/wasmtime/pull/1510#issuecomment-633693913.

bjorn3:
The branch depends on https://github.com/bytecodealliance/wasmtime/pull/1670 and an update to `object 0.19.0`.

FireFlyForLife:
Wow amazing work, I will check this out soon!

bjorn3:
The branch now also depends on https://github.com/bytecodealliance/wasmtime/pull/1767 for the `object 0.19.0` update. (You need to merge both PR's into one branch)

bjorn3:
[...]

PR #1284, opened 2022-09-27
Add Windows CI

afonso360:
👋 Hey,

With the upgrade to Cranelift 0.88 we can now enable MSVC in the build script and add some CI. This also adds CI for MinGW on Windows, which I didn't expect to hit so many issues, but it turns out MinGW is even more broken than I thought.

A summary of issues / tests disabled:

* The JIT is still broken, so we disable all JIT tagged tests on windows
* `bench.simple-raytracer`
  * This is disabled due to hyperfine failing with an invalid preparation command
     Its just a build system issue.
* `test.regex-shootout-regex-dna`
  * This fails with a different output than expected
    The actual file is mostly the same, I think this is just different line endings
* `abi-cafe` patch
  * We've dropped the previous patch, in favour of the default config for disabling the i128 windows abi issues
  But it looks like its not broad enough since `x86_64-pc-windows-gnu` still fails, so we temporarily disable those
  I'm going to follow up with  `abi-cafe` to disable those tests on MinGW as well
* `rand` patch
  * There are a bunch of tests failing on rand, these only affect  `x86_64-pc-windows-gnu` presumably due to some broken implementations on the MinGW side? I haven't investigated this too far.

cc #977


patches/0001-rand-Disable-rand-tests-on-mingw.patch
```diff
@@ -0,0 +1,47 @@
...
+diff --git a/rand_distr/src/pareto.rs b/rand_distr/src/pareto.rs
+index 217899ed9a..5f8e2c9bbb 100644
+--- a/rand_distr/src/pareto.rs
++++ b/rand_distr/src/pareto.rs
+@@ -107,6 +107,8 @@ mod tests {
+     }
+
+     #[test]
++    // This is broken on x86_64-pc-windows-gnu presumably due to a broken powf implementation
++    #[cfg(not(all(target_os = "windows", target_env = "gnu")))]
```
bjorn3, line 20: Maybe instead add `#[ignore]` on MinGW?

.github/workflows/main.yml
```diff
@@ -180,5 +202,5 @@ jobs:
     - name: Upload prebuilt cg_clif
       uses: actions/upload-artifact@v2
       with:
-        name: cg_clif-${{ runner.os }}
+        name: cg_clif-${{ runner.os }}-${{ matrix.env.TARGET_TRIPLE }}
```
bjorn3, line 214: Could you change this to just the target triple for all artifact uploads? Except for the `cg_clif-ubuntu-latest-cross-x86_64-mingw` artifact upload.

patches/0001-rand-Disable-rand-tests-on-mingw.patch
bjorn3: And could you please use 0003-... as patch name? I try not to reuse patch numbers after patches have been dropped.

.github/workflows/main.yml
afonso360: So `name: ${{ matrix.env.TARGET_TRIPLE }}` ? I'm not sure that's going to work, we only specify `TARGET_TRIPLE` in cross compilation scenarios, so the artifact table currently looks like this:

* `cg_clif-Linux`
* `cg_clif-Linux-cross-x86_64-mingw`
* `cg_clif-Windows-`
* `cg_clif-Windows-x86_64-pc-windows-gnu`
* `cg_clif-macOS`

Won't there be a conflict if we do that?

On a separate note, just noticed there is a trailing - on the Windows artifact.

bjorn3: > we only specify TARGET_TRIPLE in cross compilation scenarios.

Specifying it in non-cross compilation scenarios is fine. It defaults to the host triple, so if you set it to the host triple it doesn't affect anything.

.github/workflows/main.yml
```diff
@@ -112,7 +116,7 @@ jobs:
       if: matrix.env.TARGET_TRIPLE != 'x86_64-pc-windows-gnu'
       uses: actions/upload-artifact@v2
       with:
-        name: cg_clif-${{ runner.os }}
+        name: ${{ matrix.env.TARGET_TRIPLE }}
```
bjorn3, line 119:
```suggestion
        name: cg_clif-${{ matrix.env.TARGET_TRIPLE }}
```

.github/workflows/main.yml
```diff
@@ -211,5 +217,5 @@ jobs:
     - name: Upload prebuilt cg_clif
       uses: actions/upload-artifact@v2
       with:
-        name: cg_clif-${{ runner.os }}-${{ matrix.env.TARGET_TRIPLE }}
+        name: ${{ matrix.env.TARGET_TRIPLE }}
```
bjorn3, line 220:
```suggestion
        name: cg_clif-${{ matrix.env.TARGET_TRIPLE }}
```

bjorn3 (approved):
Thanks! Will merge once CI passes.

PR #1293, opened 2022-11-05
Update `target-lexicon` to fix new targets parsing

mati865:
This pulls in https://github.com/bytecodealliance/target-lexicon/pull/82
Noticed that `./x.py c` fails in Rust repo when using `*-gnullvm` targets because of `target-lexicon` and it could be easily fixed. Haven't tested how many further changes will be required,


bjorn3:
Thanks!

PR #1333, opened 2022-12-26
rustc_info: fix for non-rustup unix-based installations

hexagonal-sun:
If the user has installed via a package manager then rustup will not be in the user's path.  Call out to 'sh', using the `which` command to find the location of the various rust tools that are required.

### DIFF
--- a/build_system/rustc_info.rs
+++ b/build_system/rustc_info.rs
@@ -23,34 +23,32 @@ pub(crate) fn get_host_triple() -> String {
         .to_owned()
 }
 
-pub(crate) fn get_cargo_path() -> PathBuf {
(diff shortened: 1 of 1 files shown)


bjorn3:
As explained in https://github.com/bjorn3/rustc_codegen_cranelift/issues/1308#issuecomment-1351916857 using a distro provided rust version isn't going to work. cg_clif needs a specific rustc nightly as specified in rust-toolchain. I should probably add a note about this though as you are the second person in two weeks to hit this.

hexagonal-sun:
Ah, apologies for the noise.  I'll close this issue in that case.

PR #1355, opened 2023-02-10
clippy::perf fixes

matthiaskrgr:


src/concurrency_limiter.rs
```diff
@@ -19,7 +19,7 @@ impl ConcurrencyLimiter {
         let available_token_condvar = Arc::new(Condvar::new());
 
         let state_helper = state.clone();
-        let available_token_condvar_helper = available_token_condvar.clone();
+        let available_token_condvar_helper = available_token_condvar;
```
bjorn3, line 22: Something when wrong when I introduced parallel compilation. `available_token_condvar` should have been moved into `ConcurrencyLimiter` after `available_token_condvar_helper` has been moved to the helper thread, but for some reason a new Condvar is stored in `ConcurrencyLimiter`. Fixed in 4a1c2d9c7c5cb370eea0df964d537b60d29a4191. Can you please rebase and revert this change?

matthiaskrgr: :+1:  done

bjorn3:
Thanks!

PR #1356, opened 2023-02-10
clippy::complexity fixes

matthiaskrgr:


src/debuginfo/emit.rs
```diff
@@ -113,7 +113,7 @@ impl Writer for WriterRelocate {
                     offset: offset as u32,
                     size,
                     name: DebugRelocName::Symbol(symbol),
-                    addend: addend as i64,
+                    addend: addend,
```
bjorn3, line 116:
```suggestion
                    addend,
```

src/intrinsics/simd.rs
```diff
@@ -664,7 +664,7 @@ pub(super) fn codegen_simd_intrinsic_call<'tcx>(
             let m = m.load_scalar(fx);
 
             for lane in 0..lane_count {
-                let m_lane = fx.bcx.ins().ushr_imm(m, u64::from(lane) as i64);
```
bjorn3, line 667: This one is intentional to indicate the from type.

matthiaskrgr: updated and rebased

PR #1362, opened 2023-03-15
Use patched git-subtree from bjorn3/git@tqc-subtree-portable

bjorn3:
This patched has been necessary for subtree syncs from the start, but previously it was necessary to locally install tqc's patched git version, which hasn't been updated for quite a while. I made a small change to allow downloading it as script without requiring an entire git installation for the patched version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ perf.data.old
 /dist
 /rust
 /download
+/git-fixed-subtree.sh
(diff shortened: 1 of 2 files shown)


bjorn3:
To ensure the commit stays alive: https://github.com/bjorn3/git/commit/0c8268b02c29d859ed1529e08e3adb37e7ba7e38

PR #1416, opened 2023-11-04
Implement AArch64 intrinsics necessary for simd-json

afonso360:
👋 Hey,

This PR Implements the AArch64 intrinsics necessary to compile and run the simd-json testsuite.

simd-json was technically only missing `uqsub`, `addp` and I think one of the min/max ones, but I figured I might as well add the rest of them since they were quite easy.


bjorn3:
Somehow I entirely missed this PR. Thanks a lot for implementing these!

bjorn3:
> Somehow I entirely missed this PR.

It looks github removed my "watch" status for this repo. I would guess this happened when moving it to the rust-lang org. I'm now watching this repo again.

PR #1421, opened 2023-11-09
Add Rustup installation instructions to README

Kobzol:
Now that cranelift is distributed with Rustup, I find myself constantly searching for the https://github.com/rust-lang/rust/pull/81746 PR. I think that it would be useful to also describe the Rustup installation/usage instructions here in README.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -5,7 +5,47 @@ This has the potential to improve compilation times in debug mode.
 If your project doesn't use any of the things listed under "Not yet supported", it should work fine.
 If not please open an issue.
 
+## Download using Rustup
+
+The Cranelift codegen backend is distributed in nightly builds on Linux and x86_64 macOS. If you want to
+install it using Rustup, you can do that by running:
+
+```bash
+$ rustup component add rustc-codegen-cranelift-preview --toolchain nightly
+```
+
+Once it is installed, you can enable it with one of the following approaches:
+- `CARGO_PROFILE_DEV_CODEGEN_BACKEND=cranelift cargo +nightly build -Zcodegen-backend`
+- `RUSTFLAGS="-Zcodegen-backend=cranelift" cargo +nightly build`
+- Add the following to `.cargo/config.toml`:
+    ```toml
+    [unstable]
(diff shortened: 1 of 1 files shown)


lqd:
https://bjorn3.github.io/2023/10/31/progress-report-oct-2023.html describes the way to use the backend via `Cargo.toml` in order to not change the compilation commands (and also via cargo's env vars), which could be interesting here as well.

lqd:
Maybe we should also mention it in https://nnethercote.github.io/perf-book/compile-times.html

lqd:
And also document this in the unstable book.

Kobzol:
Added more usage instructions based on the blog post. Also sent PRs to the perf book (https://github.com/nnethercote/perf-book/pull/72).

I'm not sure if the unstable book is the correct place for this, as it contains just general descriptions of compiler flags. I think that the Rustc book might be more appropriate.

Readme.md
```diff
@@ -5,6 +5,34 @@ This has the potential to improve compilation times in debug mode.
 If your project doesn't use any of the things listed under "Not yet supported", it should work fine.
 If not please open an issue.
 
+## Download using Rustup
+The Cranelift codegen backend is now distributed in nightly builds on Linux. If you want to
```
bjorn3, line 9: and x86_64 macOS

bjorn3:
The actual text of the new section looks good to me. It feels a bit weird to have using precompiled versions from rustup first, then manual building and finally using precompiled versions from CI though.

Readme.md
Kobzol: Thanks, fixed.

Kobzol:
I agree. I think that it makes sense to have the easiest way first though. What order would you prefer?

bjorn3:
Maybe swap building it yourself and using the CI artifacts? You would need to reword using the CI artifacts for that as it is currently written as alternative for building yourself.

Kobzol:
I tried to reorganize it a little bit.

bjorn3:
Thanks!

Issue #1410, opened 2023-11-01
Unimplemented llvm intrinsics `aesimc` `aesdec` `aesdeclast` (used by `aes` and `ring`)


AppleSheeple:
```
warning: unsupported x86 llvm intrinsic llvm.x86.aesni.aesimc; replacing with trap
warning: unsupported x86 llvm intrinsic llvm.x86.aesni.aesdec; replacing with trap
warning: unsupported x86 llvm intrinsic llvm.x86.aesni.aesdeclast; replacing with trap
```

Issue #1409, opened 2023-10-31
Unimplemented intrinsic core_arch::x86::aes::_mm_aesdec_si128


pacak:
I'm not sure what crate it is specifically, there's about 800 dependencies but I can try to find it if it helps.

```
trap at Instance { def: Item(DefId(1:14934 ~ core[53bd]::core_arch::x86::aes::_mm_aesdec_si128)), args: [] } (_ZN4core9core_arch3x863aes16_mm_aesdec_si12817hb98140deb8307ef8E): llvm.x86.aesni.aesdec
```

PR #1425, opened 2023-11-11
Implement AES-NI and SHA256 crypto intrinsics using inline asm

bjorn3:
Fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1409
Fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1410
Part of https://github.com/rust-lang/rustc_codegen_cranelift/issues/171
cc https://github.com/RustCrypto/block-ciphers/issues/391

### DIFF
--- a/src/abi/mod.rs
+++ b/src/abi/mod.rs
@@ -383,6 +383,7 @@ pub(crate) fn codegen_terminator_call<'tcx>(
                 args,
                 ret_place,
                 target,
+                source_info.span,
             );
             return;
         }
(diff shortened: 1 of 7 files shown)


bjorn3:
Added sha256 intrinsic support too. Github has started using AMD processors with support for these instructions which was randomly causing the rustc tests to fail as bootstrap uses them when available.

benwis:
Ooh! Hoping this gets into nightly soon, nice work!

bjorn3:
Will be available in tomorrow's nightly.

PR #1432, opened 2023-11-27
Add Windows wrappers for ./y

simonbuchan:
Just a little bit to make windows dev a bit nicer. This makes `.\y ...` work in powershell and cmd, and `y ...` work in cmd.

I assume others are using git bash or something?


simonbuchan:
There's *probably* a nicer way to do exit code handling in powershell, but it's a surprisingly messy problem: https://stackoverflow.com/questions/57468522/powershell-and-process-exit-codes

bjorn3:
Another option would be to do the equivalent of `cargo build --manifest-path build_system/Cargo.toml -- "$@"` in bash. This will handle errors without any work from cmd.exe or powershell.

simonbuchan:
If you can use an MSYS bash (e.g. Git for Windows'), then `./y.sh` should work. But by default `bash` will run WSL in windows which doesn't work (performance aside it will not find `cargo.exe` when looking for `cargo`), and I think by default Git for Windows will not put it's bash on the path.

bjorn3:
I mean you could translate `cargo build --manifest-path build_system/Cargo.toml -- "$@"` from bash syntax to cmd and powershell syntax.

simonbuchan:
Well the issue is not running the built binary after building it, and ideally returning the exit code from the wrapper script. It doesn't matter if that's from rustc or cargo at that point.

Do you mean cargo *run* --manifest-path ...?

bjorn3:
> Do you mean cargo run --manifest-path ...?

Yes

simonbuchan:
Right, though I'd prefer if y.sh matched, and you would still want the wrapper to return the exit code too, so it's only a minor improvement.

PR #1443, opened 2024-01-02
Restructure x86 signed pack instructions

Nilstrieb:
This reduces the amount of duplicated code and the chance for bugs.

Sorta replaces #1442. I'm not _very_ happy with how it turned out, but it looks mostly decent I think.

Also adds new support for `_mm_packs_epi16`, `_mm256_packs_epi16` and `_mm256_packus_epi32`.

For reference, here's a table of all the instructions:

|          | sse 16                                       | avx 16                                      | sse 32                                      | avx 32                                      |
|----------|----------------------------------------------|---------------------------------------------|---------------------------------------------|---------------------------------------------|
| unsigned | _mm_packus_epi16\|llvm.x86.sse2.packuswb.128 | _mm256_packus_epi16\|llvm.x86.avx2.packuswb | _mm_packus_epi32\|llvm.x86.sse41.packusdw   | _mm256_packus_epi32\|llvm.x86.avx2.packusdw |
| signed   | _mm_packs_epi16\|llvm.x86.sse2.packsswb.128  | _mm256_packs_epi16\|llvm.x86.avx2.packsswb  | _mm_packs_epi32\|llvm.x86.sse2.packssdw.128 | _mm256_packs_epi32\|llvm.x86.avx2.packssdw  |

### DIFF
--- a/src/intrinsics/llvm_x86.rs
+++ b/src/intrinsics/llvm_x86.rs
@@ -610,230 +610,56 @@ pub(crate) fn codegen_x86_llvm_intrinsic_call<'tcx>(
             // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_packus_epi16&ig_expand=4903
             intrinsic_args!(fx, args => (a, b); intrinsic);
 
-            assert_eq!(a.layout(), b.layout());
-            let layout = a.layout();
+            pack_instruction(fx, a, b, ret, PackSize::U8, PackWidth::Sse);
+        }
 
-            let (lane_count, lane_ty) = layout.ty.simd_size_and_type(fx.tcx);
-            let (ret_lane_count, ret_lane_ty) = ret.layout().ty.simd_size_and_type(fx.tcx);
-            assert_eq!(lane_ty, fx.tcx.types.i16);
-            assert_eq!(ret_lane_ty, fx.tcx.types.u8);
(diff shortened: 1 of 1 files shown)


src/intrinsics/llvm_x86.rs
```diff
@@ -1407,3 +1297,106 @@ fn llvm_add_sub<'tcx>(
...
+    fn ret_ty<'tcx>(&self, tcx: TyCtxt<'tcx>) -> Ty<'tcx> {
+        match self {
+            Self::U8 => tcx.types.u8,
+            Self::U16 => tcx.types.u16,
+        }
+    }
+    fn max(&self) -> i64 {
+        match self {
+            Self::U8 => u8::MAX as u64 as _,
+            Self::U16 => u16::MAX as u64 as _,
```
bjorn3, line 1334: This should be i8/i16::MAX for the signed output case, right? And similarly i8/i16::MIN for the second argument to smax.

Nilstrieb: You're right. There were also a lot of other subtle bugs in the code I put up yesterday. To fix all of those bugs and ensure the new implementation is correct, I compared a bunch of random input against LLVM and was finally able to get it working. See commit c8f5d35508e062bd2d95e6c03429bfec831db6d3.

bjorn3:
Thanks!

PR #1444, opened 2024-01-03
Use `rust-analyzer.rustc.source` to get r-a working with rustc

Nilstrieb:


.vscode/settings.json
```diff
@@ -1,8 +1,7 @@
 {
     "editor.formatOnSave": true,
 
-    // source for rustc_* is not included in the rust-src component; disable the errors about this
-    "rust-analyzer.diagnostics.disabled": ["unresolved-extern-crate", "unresolved-macro-call"],
```
bjorn3, line 5: Can you keep this config to make it easier to comment out the `rust-analyzer.rustc.source` if necessary for performance reasons?

PR #1462, opened 2024-03-03
debuginfo: embed gdb visualizers in module

### LINKED ISSUE #1386: Implement #![debugger_visualizer]
https://doc.rust-lang.org/nightly/reference/attributes/debugger.html#using-debugger_visualizer-with-gdb
https://github.com/rust-lang/rust/blob/ab0f3e680e4fd50218914b8afa6900a67377721c/compiler/rustc_codegen_llvm/src/debuginfo/gdb.rs

The `.debug_gdb_scripts` section should contain `\x01gdb_load_rust_pretty_printers.py` followed by for each `#![debugger_visualizer(gdb_script_file = "...")]` the pretty printer inlined into the executable. Except if one of the conditions to omit `.debug_gdb_scripts` is true.

kartva:
fixes #1386

### DIFF
--- a/src/debuginfo/emit.rs
+++ b/src/debuginfo/emit.rs
@@ -35,6 +35,23 @@ impl DebugContext {
             }
             Ok(())
         });
+
+        self.add_gdb_debugger_visualizers(product);
+    }
+
+    fn add_gdb_debugger_visualizers(&mut self, product: &mut ObjectProduct) {
+        let mut section = WriterRelocate::new(self.endian);
+        section.write(b"\x01gdb_load_rust_pretty_printers.py\0").unwrap();
+
+        for (index, visualizer) in self.gdb_visualizers.iter().enumerate() {
+            let vis_name = format!("\x04pretty-printer-{}-{index}\n", self.crate_name);
+
+            section.write(vis_name.as_bytes()).unwrap();
+            section.write(&visualizer.src).unwrap();
+
+            section.write(b"\0").unwrap();
+        }
+        product.add_gdb_visualizers_section(section.writer.take());
     }
 }
 
--- a/src/debuginfo/mod.rs
+++ b/src/debuginfo/mod.rs
@@ -5,6 +5,8 @@ mod line_info;
 mod object;
 mod unwind;
 
+use std::collections::BTreeSet;
+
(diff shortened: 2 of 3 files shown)


src/debuginfo/emit.rs
```diff
@@ -35,6 +35,23 @@ impl DebugContext {
             }
             Ok(())
         });
+
+        self.add_gdb_debugger_visualizers(product);
```
bjorn3, line 39: This function needs to be called at https://github.com/rust-lang/rustc_codegen_cranelift/blob/8c46d93ca0f287ea3b284c061eebcb4a275fc43a/src/main_shim.rs#L40 and a load of the produced global needs to be inserted there too to ensure the section only gets inserted a single time in the whole program and that it doesn't get removed by the linker.

src/debuginfo/emit.rs
```diff
@@ -35,6 +35,23 @@ impl DebugContext {
...
+    }
+
+    fn add_gdb_debugger_visualizers(&mut self, product: &mut ObjectProduct) {
```
bjorn3, line 42: If https://github.com/rust-lang/rust/blob/7606c13961ddc1174b70638e934df0439b7dc515/compiler/rustc_codegen_llvm/src/debuginfo/gdb.rs#L85-L117 returns false you need to skip adding the section.

src/debuginfo/emit.rs
```diff
@@ -35,6 +35,23 @@ impl DebugContext {
...
+        let mut section = WriterRelocate::new(self.endian);
```
bjorn3, line 43: No need to go through `WriterRelocate` here. A plain `Vec` is enough. There will be no relocations inside this section.

ChiefMateStarbuck:
Is this PR still needed? I can pick this up and clean it up.

bjorn3:
Yes, feel free to pick it up.

kernel-loophole:
@ChiefMateStarbuck are you working on it?

ChiefMateStarbuck:
@kernel-loophole , hello. I am not currently working on this. You are free to pick it up.

kernel-loophole:
thanks @ChiefMateStarbuck

PR #1467, opened 2024-03-21
Avoid ICE when global_asm const operand fails to evaluate

### LINKED ISSUE #1466: Avoid ICE when global_asm const operand fails to evaluate
See https://github.com/rust-lang/rust/pull/122691

BeetleFunk:
Fixes #1466 

Follows the same approach used in rust-lang/rust#122691. The error is reported when calling `const_eval_poly()`. `codegen_global_asm_item()` doesn't have a way to propagate the error but can detect the `ErrorHandled::Reported` case and avoid the panic.

### DIFF
--- a/src/global_asm.rs
+++ b/src/global_asm.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 
 use rustc_ast::{InlineAsmOptions, InlineAsmTemplatePiece};
 use rustc_hir::{InlineAsmOperand, ItemId};
+use rustc_middle::mir::interpret::ErrorHandled;
 use rustc_session::config::{OutputFilenames, OutputType};
 use rustc_target::asm::InlineAsmArch;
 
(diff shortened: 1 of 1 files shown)


BeetleFunk:
With these changes, compiling the [test case file:](https://github.com/rust-lang/rust/blob/394821060dac473390d723473217f0b2d630fc44/tests/ui/asm/fail-const-eval-issue-121099.rs)
```
//@ build-fail
//@ needs-asm-support
#![feature(asm_const)]

use std::arch::global_asm;

fn main() {}

global_asm!("/* {} */", const 1 << 500); //~ ERROR evaluation of constant value failed [E0080]

global_asm!("/* {} */", const 1 / 0); //~ ERROR evaluation of constant value failed [E0080]
```
Results in STDERR output:
```
error[E0080]: evaluation of constant value failed
 --> src/main.rs:9:31
  |
9 | global_asm!("/* {} */", const 1 << 500); //~ ERROR evaluation of constant value failed [E0080]
  |                               ^^^^^^^^ attempt to shift left by `500_i32`, which would overflow

error[E0080]: evaluation of constant value failed
  --> src/main.rs:11:31
   |
11 | global_asm!("/* {} */", const 1 / 0); //~ ERROR evaluation of constant value failed [E0080]
   |                               ^^^^^ attempt to divide `1_i32` by zero

For more information about this error, try `rustc --explain E0080`.
error: could not compile `sandbox` (bin "sandbox") due to 2 previous errors
```

BeetleFunk:
What's standard practice on automated tests for compiler errors like this?

I noticed that the test runner in `scripts/test_rustc_tests.sh` seems to be filtering out most of the rustc test files containing "build-fail" or "ERR" patterns.

bjorn3:
Thanks!

Issue #1487, opened 2024-05-09
Support compiling zlib-rs


bjorn3:
https://github.com/memorysafety/zlib-rs

Missing intrinsics:

* [x] llvm.x86.sse42.crc32.32.32 (https://github.com/rust-lang/rustc_codegen_cranelift/pull/1488)
* [x] llvm.x86.avx2.permd (https://github.com/rust-lang/rustc_codegen_cranelift/pull/1491)

PR #1488, opened 2024-05-11
add the `llvm.x86.sse42.crc32.32.32` intrinsic

folkertdev:
part of #1487 

how should this be tested? I exercised the instruction by compiling and running (part of) the test suite of zlib-rs, is that sufficient?

I can add the other variations of this instruction (having a differently-sized `v` argument) but those are not used by zlib-rs so some dedicated testing would presumably be required for them?

### DIFF
--- a/src/intrinsics/llvm_x86.rs
+++ b/src/intrinsics/llvm_x86.rs
@@ -832,6 +832,32 @@ pub(crate) fn codegen_x86_llvm_intrinsic_call<'tcx>(
             }
         }
 
+        "llvm.x86.sse42.crc32.32.32" => {
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#ig_expand=939,1419,1419&text=_mm_crc32_u32
+            intrinsic_args!(fx, args => (crc, v); intrinsic);
+
+            let crc = crc.load_scalar(fx);
+            let v = v.load_scalar(fx);
+
+            codegen_inline_asm_inner(
+                fx,
+                &[InlineAsmTemplatePiece::String("crc32 eax, edx".to_string())],
+                &[
+                    CInlineAsmOperand::InOut {
+                        reg: InlineAsmRegOrRegClass::Reg(InlineAsmReg::X86(X86InlineAsmReg::ax)),
+                        _late: true,
+                        in_value: crc,
+                        out_place: Some(ret),
+                    },
+                    CInlineAsmOperand::In {
+                        reg: InlineAsmRegOrRegClass::Reg(InlineAsmReg::X86(X86InlineAsmReg::dx)),
+                        value: v,
+                    },
+                ],
+                InlineAsmOptions::NOSTACK | InlineAsmOptions::PURE | InlineAsmOptions::NOMEM,
+            );
+        }
+
         "llvm.x86.sse42.pcmpestri128" => {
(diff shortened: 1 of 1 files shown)


src/intrinsics/llvm_x86.rs
```diff
@@ -832,6 +832,32 @@ pub(crate) fn codegen_x86_llvm_intrinsic_call<'tcx>(
...
+                    CInlineAsmOperand::InOut {
+                        reg: InlineAsmRegOrRegClass::Reg(InlineAsmReg::X86(X86InlineAsmReg::ax)),
+                        _late: true,
+                        in_value: crc,
+                        out_place: Some(ret),
+                    },
+                    CInlineAsmOperand::In {
+                        reg: InlineAsmRegOrRegClass::Reg(InlineAsmReg::X86(X86InlineAsmReg::dx)),
+                        value: v,
+                    },
```
folkertdev, line 855: is there any particular reason to choose particular register names here? this seemed to be the ordering used in some of the other examples.

src/intrinsics/llvm_x86.rs
```diff
@@ -832,6 +832,32 @@ pub(crate) fn codegen_x86_llvm_intrinsic_call<'tcx>(
             }
         }
 
+        "llvm.x86.sse42.crc32.32.32" => {
+            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#ig_expand=939,1419,1419&text=_mm_crc32_u32
```
bjorn3, line 836:
```suggestion
            // https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#ig_expand=1419&text=_mm_crc32_u32
```

nit

src/intrinsics/llvm_x86.rs
bjorn3: There is no particular reason for the choice of registers.

bjorn3 (commented):
> I can add the other variations of this instruction (having a differently-sized v argument) but those are not used by zlib-rs so some dedicated testing would presumably be required for them?

Running the stdarch test suite should cover this. I can run it for you if you want add the other variants, but implementing them as needed is fine too.

> how should this be tested? I exercised the instruction by compiling and running (part of) the test suite of zlib-rs, is that sufficient?

`example/std_example.rs` has various simd tests. You can copy the test from stdarch there and call it as regular function in `test_simd`.

bjorn3:
Thanks!

folkertdev:
that testing approach does not quite work because

```
[AOT] arbitrary_self_types_pointers_and_wrappers
[BUILD] alloc_system
[AOT] alloc_example
Hello World!
[JIT] std_example
error: Inline asm is not supported in JIT mode

error: aborting due to 1 previous error
```

so then I guess local testing just has to suffice (for now)? 

for future reference, here is the test

```rust
#[cfg(target_arch = "x86_64")]
#[target_feature(enable = "sse4.2")]
unsafe fn test_crc32() {
    assert!(is_x86_feature_detected!("sse4.2"));

    let a = 42u32;
    let b = 0xdeadbeefu64;

    assert_eq!(_mm_crc32_u8(a, b as u8), 4135334616);
    assert_eq!(_mm_crc32_u16(a, b as u16), 1200687288);
    assert_eq!(_mm_crc32_u32(a, b as u32), 2543798776);
    assert_eq!(_mm_crc32_u64(a as u64, b as u64), 241952147);
}
```

bjorn3:
Adding `#[cfg(not(jit))]` should work.

PR #1490, opened 2024-05-11
add all `llvm.x86.sse42.crc32.*.*` intrinsics

folkertdev:
followup  to https://github.com/rust-lang/rustc_codegen_cranelift/pull/1488


example/std_example.rs
```diff
@@ -122,6 +122,9 @@ fn main() {
     #[cfg(target_arch = "x86_64")]
     unsafe {
         test_simd();
+
+        #[cfg(not(jit))]
+        test_crc32();
```
bjorn3, line 127: Can you move this call into `test_simd`?

Issue #1451, opened 2024-01-26
arm64 macOS variadic arguments ABI needs Cranelift support


bjorn3:
On pretty much every target variadic functions almost the exact same ABI (on x86 you need to set `%al` to the upper bound of the amount of float args, so float args are currently forbidden by cg_clif) as non-variadic functions with the same arguments passed (after integer promotion, but rust denies vararg calls that need integer promotion anyway). This means that I can work around the lack of vararg support in Cranelift by changing the function signature of the imported function declaration to match the non-variadic counterpart. On arm64 macOS however variadic arguments are passed differently from non-variadic arguments: https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Update-code-that-passes-arguments-to-variadic-functions As such we can't support vararg functions on arm64 macOS until Cranelift gets native vararg support: https://github.com/bytecodealliance/wasmtime/issues/1030

PR #1500, opened 2024-06-18
Fix varargs support on `aarch64-apple-darwin`

beetrees:
Uses the strategy outlined in [#1451 (comment)](https://github.com/rust-lang/rustc_codegen_cranelift/issues/1451#issuecomment-2169248769) to enable support for varargs functions with integer/pointer arguments on `aarch64-apple-darwin`. CI testing is also added for `aarch64-apple-darwin`.

### DIFF
--- a/src/abi/mod.rs
+++ b/src/abi/mod.rs
@@ -5,8 +5,9 @@ mod pass_mode;
 mod returning;
 
 use std::borrow::Cow;
+use std::mem;
 
-use cranelift_codegen::ir::SigRef;
+use cranelift_codegen::ir::{ArgumentPurpose, SigRef};
 use cranelift_codegen::isa::CallConv;
 use cranelift_module::ModuleError;
 use rustc_codegen_ssa::errors::CompilerBuiltinsCannotCall;
@@ -17,7 +18,7 @@ use rustc_middle::ty::TypeVisitableExt;
 use rustc_monomorphize::is_call_from_compiler_builtins_to_upstream_monomorphization;
 use rustc_session::Session;
 use rustc_span::source_map::Spanned;
-use rustc_target::abi::call::{Conv, FnAbi};
+use rustc_target::abi::call::{Conv, FnAbi, PassMode};
 use rustc_target::spec::abi::Abi;
 
 use self::pass_mode::*;
@@ -487,6 +488,7 @@ pub(crate) fn codegen_terminator_call<'tcx>(
     let args = args;
     assert_eq!(fn_abi.args.len(), args.len());
 
+    #[derive(Copy, Clone)]
     enum CallTarget {
         Direct(FuncRef),
         Indirect(SigRef, Value),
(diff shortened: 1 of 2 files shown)


src/abi/mod.rs
```diff
@@ -532,18 +533,64 @@ pub(crate) fn codegen_terminator_call<'tcx>(
...
-                        adjust_arg_for_abi(fx, arg.value, &fn_abi.args[i], arg.is_owned).into_iter()
-                    }),
-            )
-            .collect::<Vec<Value>>();
+        let call_args = if fx.tcx.sess.target.is_like_osx
+            && fx.tcx.sess.target.arch == "aarch64"
+            && fn_sig.c_variadic()
+            // No need to pad the argument list unless vardiac arguments are actually being passed.
+            && args.len() > fixed_arg_count
+        {
```
bjorn3, line 541: This is a lot of code here. Would it be possible to move it to the `if fn_sig.c_variadic() {` if statement after emitting the call inst? Basically the idea would be to modify the argument list of the call instruction and splice extra arguments in. You can probably use `func.dfg.overwrite_inst_values`, taking care to account for the first arg of `call_indirect` being the function pointer rather than a call arguments. This would keep all vararg hacks in a single place which can easily be removed once Cranelift lands real vararg support.

src/abi/mod.rs
```diff
@@ -532,18 +533,64 @@ pub(crate) fn codegen_terminator_call<'tcx>(
...
+                .chain(
+                    (&mut args_iter)
+                        .take(fixed_arg_count)
+                        .skip(if first_arg_override.is_some() { 1 } else { 0 })
+                        .flat_map(|(i, arg)| {
+                            adjust_arg_for_abi(fx, arg.value, &fn_abi.args[i], arg.is_owned)
+                                .into_iter()
+                        }),
+                )
+                .collect::<Vec<Value>>();
```
bjorn3, line 556: If there is a struct arg before the variadic args does that immediately force all following args to be on the stack already? If so please either reject struct args in variadic calls or take them into account.

bjorn3:
Thanks for the PR! While this is a big hack, so is the rest of the vararg support in cg_clif, so I don't see it as a reason to reject this PR. I would like to see all parts of the vararg support hack to be placed together though. Also if you don't mind, could you ask a CI job for aarch64-apple-darwin to test it? This should be a matter of duplicating the block at https://github.com/rust-lang/rustc_codegen_cranelift/blob/1cb728280dfe7cc241eae34196f80c876aa9addc/.github/workflows/main.yml#L57-L59 and adjusting the `TARGET_TRIPLE`. You can also do the same at https://github.com/rust-lang/rustc_codegen_cranelift/blob/1cb728280dfe7cc241eae34196f80c876aa9addc/.github/workflows/main.yml#L214-L216 to get CI to upload a prebuilt version of cg_clif for arm64 macOS.

src/abi/mod.rs
bjorn3: I see. Maybe `match func.stencil.dfg.inst[call_inst] { InstructionData::Call { args } => { ... }, InstructionData::CallIndirect { args, ... } => { ... }` with `args.insert(index, zero, &mut func.stencil.dfg.value_lists)` to actually add the arguments would work? If not what you have right now is fine as is.

bjorn3:
@beetrees I want to post an announcement on Mastodon. Do you want me to credit you as @\beetrees?

bjorn3:
https://hachyderm.io/@bjorn3/112648887843960881

Issue #1511, opened 2024-07-13
Use minimal rustup profile in CI


bjorn3:
This should make downloading the toolchain a bit faster. Especially on Windows where it currently takes about 2 minutes. Setting `profile = "minimal"` in `rust-toolchain` may work, but otherwise running `rustup set profile minimal` before `./y.sh prepare` should do the trick I think.

PR #1514, opened 2024-07-14
Set rustup profile to minimal thru rustup #1511

vidurkhanal:
Added a command on CI run to set profile to minimal using rustup.

Addreses #1511

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,7 +113,7 @@ jobs:
 
     - name: Prepare dependencies
       run: |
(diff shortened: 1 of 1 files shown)


bjorn3:
Thanks for the PR! Unfortunately as it turns out `rustup set profile minimal` doesn't work. It still downloads clippy and such in CI. I expected it to work, but turns out I was wrong about this. In the mean time someone else unaware of your attempts opened a PR that did actually got a fix for https://github.com/rust-lang/rustc_codegen_cranelift/issues/1511 working. I've merged that PR instead. Thanks for attempting to fix this issue anyway!

PR #1516, opened 2024-07-14
Set rustup profile to minimal

### LINKED ISSUE #1511: Use minimal rustup profile in CI

lqd:
This PR fixes #1511 by setting the minimal profile in the toolchain file, and removing the explicit places where this was already done in CI scripts.

I don't think using `rustup set profile minimal` as you mention in the issue works when there's already a rustup override file: I believe the override kinda sets the profile back to its default value.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -13,7 +13,6 @@ jobs:
       - uses: actions/checkout@v4
       - run: |
           sed -i 's/components.*/components = []/' rust-toolchain
-          echo 'profile = "minimal"' >> rust-toolchain
       - uses: rustsec/audit-check@v1.4.1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


lqd:
Ah I didn't notice there was [an open PR](https://github.com/rust-lang/rustc_codegen_cranelift/pull/1514) to fix the same issue. But since it's using the `rustup set profile minimal` trick, we can verify my assumption that it wouldn't work.

In that PR, for example in the [`ubuntu-20.04` dist job](https://github.com/rust-lang/rustc_codegen_cranelift/actions/runs/9924689051/job/27416270698?pr=1514#step:7:12), components that are not in the minimal profile are still being downloaded: 

> info: downloading component 'rust-docs'

Compared to using the `rust-toolchain` file, in this PR's [same job](https://github.com/rust-lang/rustc_codegen_cranelift/actions/runs/9926416155/job/27419976685?pr=1516#step:7:10) which does look to be correctly set to "minimal".

PR #1519, opened 2024-07-20
Remove unneeded libcalls from `codegen_i128.rs`

beetrees:
All 4 Cranelift backends now support 128-bit multiplication, so the libcall for it can be removed. Additionally, the checked add/sub implementation in [`num.rs`](https://github.com/rust-lang/rustc_codegen_cranelift/blob/b929b68a5ba4570e15637b12d9e687a2ba6673d8/src/num.rs#L210-L237) works for 128-bit integers, so the libcalls for them can also be removed.

### DIFF
--- a/src/codegen_i128.rs
+++ b/src/codegen_i128.rs
@@ -23,19 +23,7 @@ pub(crate) fn maybe_codegen<'tcx>(
     match bin_op {
         BinOp::BitAnd | BinOp::BitOr | BinOp::BitXor => None,
         BinOp::Add | BinOp::AddUnchecked | BinOp::Sub | BinOp::SubUnchecked => None,
-        BinOp::Mul | BinOp::MulUnchecked => {
-            let args = [lhs.load_scalar(fx), rhs.load_scalar(fx)];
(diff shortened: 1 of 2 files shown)


src/codegen_i128.rs
```diff
@@ -121,15 +110,7 @@ pub(crate) fn maybe_codegen_checked<'tcx>(
                 AbiParam::new(types::I128),
             ];
             let args = [out_place.to_ptr().get_addr(fx), lhs.load_scalar(fx), rhs.load_scalar(fx)];
-            let name = match (bin_op, is_signed) {
-                (BinOp::Add, false) => "__rust_u128_addo",
-                (BinOp::Add, true) => "__rust_i128_addo",
-                (BinOp::Sub, false) => "__rust_u128_subo",
-                (BinOp::Sub, true) => "__rust_i128_subo",
```
bjorn3, line 128: Can you update https://github.com/rust-lang/rustc_codegen_cranelift/blob/master/src/compiler_builtins.rs to remove the now unused intrinsics?

bjorn3:
Thanks!

PR #1521, opened 2024-07-21
Fix handling of large alignments in `write_cvalue_maybe_transmute`

beetrees:
The `as` cast that was used previously would lead to the alignment being truncated such that an alignment of zero would be passed to `emit_small_memory_copy`.


bjorn3:
This doesn't currently have any effect as types with larger than 128 byte alignment are emitted as memcpy anyway, which doesn't use the alignment hint. It could have an effect in the future, so I merged it anyway.

PR #1550, opened 2024-12-18
hook up tracing to cg_cranelift

jyn514:
this was easier than expected. here is an example of using RUSTC_LOG with a build of cranelift from rust-lang/rust:
```
$ RUSTC_LOG=rustc_codegen_cranelift cargo +stage1 b
   Compiling example v0.1.0 (/home/jyn/src/example)
 INFO rustc_codegen_cranelift codegen crate example
 INFO rustc_codegen_cranelift codegen crate example
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.44s
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,8 @@ extern crate rustc_metadata;
 extern crate rustc_session;
 extern crate rustc_span;
 extern crate rustc_target;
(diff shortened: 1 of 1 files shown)


bjorn3:
Thanks!

Issue #1553, opened 2025-01-11
Latest dev version (a642651) hangs forever on link.exe


frederikhors:
Tried today the latest dev version and the `cargo-clif build`` is stopping on `link.exe` forever!

I'm coming from `cargo 1.85.0-nightly (769f622e1 2024-12-14)` that works OK.

I'm on Microsoft `14.42.34433.0`, Windows 11.

bjorn3:
I'm afraid I don't have a clue what could be going on. Maybe if you can bisect it to the exact commit that introduced this issue that would help explain it? If you don't want to build cg_clif locally, you can download the CI artifacts for a specific commit from https://github.com/rust-lang/rustc_codegen_cranelift/actions/workflows/main.yml?query=branch%3Amaster If you click on the CI run for the commit in question, you can scroll down to the bottom to see the release artifacts. Github only keeps them for a relatively limited amount of time, though I expect the artifacts for 2024-12-14 to still exist.

bjorn3:
Actually maybe https://github.com/rust-lang/rustc_codegen_cranelift/commit/4cc1c9036d01f0ee2077e2f04a388bbe36424c54 is the offending commit? Can you check if it works on the commit before this and breaks with this commit?

frederikhors:
[4cc1c90](https://github.com/rust-lang/rustc_codegen_cranelift/commit/4cc1c9036d01f0ee2077e2f04a388bbe36424c54) has the issue.

I'm trying the one before now (https://github.com/rust-lang/rustc_codegen_cranelift/commit/7f98f34420e35680fb7426dae9133513135b1122).

frederikhors:
[4cc1c90](https://github.com/rust-lang/rustc_codegen_cranelift/commit/4cc1c9036d01f0ee2077e2f04a388bbe36424c54) has the issue. Plus it's very, very, very slower before arriving to the last crate (when it calls link.exe).

[I'm trying the one before now (https://github.com/rust-lang/rustc_codegen_cranelift/commit/7f98f34420e35680fb7426dae9133513135b1122).](https://github.com/rust-lang/rustc_codegen_cranelift/commit/7f98f34420e35680fb7426dae9133513135b1122) doesn't have the problem. And it works with similar timing to the commit I've been using so far.

bjorn3:
[...]

PR #1554, opened 2025-01-14
Disable -Zfunction-sections by default on MSVC

bjorn3:
To workaround a hang in link.exe.

Fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1553


bjorn3:
> is it only being disabled for msvc btw?

Yes, it was. I have changed it to disabling it on all Windows targets though as rustc disables it on MinGW already anyway.

lqd:
(Lol the msvc check was right in there I am a doofus)

PR #1556, opened 2025-01-20
Ignore global git hooks when importing rust-lang/rust git repo

jyn514:
These may cause tests to fail. Disable them.

### DIFF
--- a/build_system/prepare.rs
+++ b/build_system/prepare.rs
@@ -186,7 +186,7 @@ fn init_git_repo(repo_dir: &Path) {
     spawn_and_wait(git_add_cmd);
 
     let mut git_commit_cmd = git_command(repo_dir, "commit");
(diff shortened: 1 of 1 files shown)


build_system/prepare.rs
```diff
@@ -186,7 +186,7 @@ fn init_git_repo(repo_dir: &Path) {
     spawn_and_wait(git_add_cmd);
 
     let mut git_commit_cmd = git_command(repo_dir, "commit");
-    git_commit_cmd.arg("-m").arg("Initial commit").arg("-q");
+    git_commit_cmd.arg("-m").arg("Initial commit").arg("-q").arg("-n");
```
bjorn3, line 189: Please spell it out as `--no-verify` to make it clearer what it does.

PR #1557, opened 2025-01-21
Test and dist for arm64 linux on CI

bjorn3:


bjorn3:
I'm going to wait for https://github.com/rust-lang/rust/issues/135867 to be resolved first.

bjorn3:
Github has fixed the crashes by reverting back to an older CPU model.

PR #1565, opened 2025-03-18
Remove liballoc compiler-builtins f16/f128 disable patch

bjorn3:
We already build with the compiler-builtins-no-f16-f128 feature enabled.


bjorn3:
Didn't mean to open a PR for this.

PR #1568, opened 2025-04-08
Replace trap_unimplemented calls with codegen_panic_nounwind

bjorn3:
This will show a backtrace. Also added a reference to rust-lang/rustc_codegen_cranelift#171 in the unimplemented intrinsic error message.


caspark:
Thank you for jumping on my feedback so quickly!

Issue #1555, opened 2025-01-15
Support -Zmin-function-alignment


bjorn3:
This was added in https://github.com/rust-lang/rust/pull/134030. This will require a minor change to cranelif-object to be able to increase the alignment at https://github.com/bytecodealliance/wasmtime/blob/7f9049b9df84a748d0f7b4024e29b26014923c28/cranelift/object/src/backend.rs#L378-L380

bjorn3:
Once https://github.com/bytecodealliance/wasmtime/pull/10391 is merged this can be implemented using:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index caaa0696..bc8fba81 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -278,6 +278,12 @@ fn build_isa(sess: &Session) -> Arc<dyn TargetIsa + 'static> {
 
     flags_builder.set("enable_llvm_abi_extensions", "true").unwrap();
 
+    if let Some(min_function_alignment) = sess.opts.unstable_opts.min_function_alignment {
+        flags_builder
+            .set("log2_min_function_alignment", &min_function_alignment.bytes().ilog2().to_string())
+            .unwrap();
+    }
+
     use rustc_session::config::OptLevel;
     match sess.opts.optimize {
         OptLevel::No => {
```

PR #1572, opened 2025-04-20
support `-Zmin-function-alignment`

folkertdev:
fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1555

Based on the timing, this should be supported now 

- https://github.com/bytecodealliance/wasmtime/pull/10391 was merged on March 13th
- https://crates.io/crates/cranelift was last released on March 20th

I strongly suspect that no tests will hit this though (though I don't see how the relevant tests are ignored in `scripts/test_rustc_tests.sh`). As far as I can tell, function alignment with `#[repr(align(N))]` is not yet supported. I vaguely rember we discussed how this could be added, but now I can't find a way in the cranelift docs to align an individual function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -280,6 +280,12 @@ fn build_isa(sess: &Session, jit: bool) -> Arc<dyn TargetIsa + 'static> {
 
     flags_builder.set("enable_llvm_abi_extensions", "true").unwrap();
 
+    if let Some(align) = sess.opts.unstable_opts.min_function_alignment {
+        flags_builder
+            .set("log2_min_function_alignment", &align.bytes().ilog2().to_string())
+            .unwrap();
+    }
+
     use rustc_session::config::OptLevel;
     match sess.opts.optimize {
         OptLevel::No => {


bjorn3:
The Cranelift PR got merged after the current release branched. It will be in the next Cranelift release, which will be released somewhere ij the next couple of days.

There is a release every month and two weeks before the release, a new branch is created for it to let the changes bake a bit and have bugfixes backported without introducing new regressions.

folkertdev:
Ah, well the next release should be soon then. But also, that this passed CI means that the logic never actually gets hit right? So the relevant tests are ignored by something in `scripts/test_rustc_tests.sh`?

bjorn3:
The only tests for this are codegen tests, which don't get run by test_rustc_tests.sh as `--emit asm` is not supported by cg_clif and in general those tests are too specific in what instructions they match to pass with any backend other than the LLVM backend anyway.

bjorn3:
The master branch is now updated to the new Cranelift release.

folkertdev:
I've rebased, with that:

- should we do anything in terms of testing now?
- do you have pointers for how to support `#[repr(align(N))]` on functions? We can then add some UI tests for the flag and the attribute, just to assert that the address in practice is actually a multiple of the specified alignment.

bjorn3:
> do you have pointers for how to support #[repr(align(N))] on functions?

That can't be implemented yet until cranelift-module supports specifying a TargetIsa for individual function.

> should we do anything in terms of testing now?

If you have locally checked that it sets the correct alignment in the function sections, that is enough as far as I'm concerned.

folkertdev:
I ran this, so I think we're good

```
> cat src/main.rs
#![feature(fn_align)]

#[repr(align(512))]
#[no_mangle]
fn foo() {}

#[repr(align(512))]
fn main() {
    assert!((foo as usize).is_multiple_of(512));
    assert!((main as usize).is_multiple_of(512));
}

> RUSTFLAGS="-Zmin-function-alignment=512" ~/rust/rustc_codegen_cranelift/dist/cargo-clif run
   Compiling playground v0.1.0 (/home/folkertdev/rust/playground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/playground`
> RUSTFLAGS="-Zmin-function-alignment=256" ~/rust/rustc_codegen_cranelift/dist/cargo-clif run
   Compiling playground v0.1.0 (/home/folkertdev/rust/playground)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/playground`

thread 'main' panicked at src/main.rs:10:5:
assertion failed: (main as usize).is_multiple_of(512)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fish: Job 1, 'RUSTFLAGS="-Zmin-function-align…' terminated by signal SIGABRT (Abort)
```

folkertdev:
> That can't be implemented yet until cranelift-module supports specifying a TargetIsa for individual function.

Is there an issue for that that I'm missing?

Issue #1461, opened 2024-03-01
Support `f16` and `f128`


tgross35:
As of https://github.com/rust-lang/rust/pull/121728, the backends now have stubs for `f16` and `f128`: https://github.com/rust-lang/rust/blob/6cbf0926d54c80ea6d15df333be9281f65bbeb36/compiler/rustc_codegen_cranelift/src/common.rs#L36-L39.

It will be a while before these codepaths can even be exercised since the frontend isn't hooked up yet, but at some point this should be changed to either gracefully fail or do the right thing.

bjorn3:
This will need direct Cranelift support. Back ib the day before Cranelift had native 128bit int support I had a whole bunch of hacks to make it work. They were really fragile and I don't think they will work anymore for the current structure of cg_clif.

tgross35:
What all would be needed for clif to support these in a minimal state?

bjorn3:
At a minimum Cranelift support for the f16 and f128 values. And either Cranelift support for float instructions for these types or emulation in cg_clif using libm calls in https://github.com/rust-lang/rustc_codegen_cranelift/blob/8c46d93ca0f287ea3b284c061eebcb4a275fc43a/src/num.rs#L313-L363 and https://github.com/rust-lang/rustc_codegen_cranelift/blob/8c46d93ca0f287ea3b284c061eebcb4a275fc43a/src/intrinsics/mod.rs#L309-L349.

Nemo157:
It seems like the frontend is now hooked up and this is used in some sysroot crates like `compiler-builtins` so using `-Zbuild-std` with cranelift fails, is there some way to avoid this now?

tgross35:
`compiler-builtins` has the `no-f16-f128` feature to disable both of these https://github.com/rust-lang/compiler-builtins/blob/adaef3265c560998a03b6902bd63d88009d3a132/Cargo.toml#L55, and then std carries a flag to propagate that https://github.com/rust-lang/rust/blob/017ae1b21f7be6dcdcfc95631e54bde806653a8a/library/std/Cargo.toml#L103. Maybe there is a way to automatically set this flag based on codegen backend in build-std?

Nemo157:
Ah yes, I'd found the `compiler-builtins` feature but missed the `std` one, this works now:

```
export CARGO_UNSTABLE_BUILD_STD_FEATURES=compiler-builtins-no-f16-f128
[...]

PR #1574, opened 2025-04-22
Add `f16`/`f128` support

beetrees:
This PR adds full support for `f16` and `f128`. As Cranelift backends only fully support basic operations such as load, store and bitcast at the moment, operations are polyfilled to allow full support for everything currently implemented in `rustc`.

Fixes rust-lang/rustc_codegen_cranelift#1461

### DIFF
--- a/src/codegen_f16_f128.rs
+++ b/src/codegen_f16_f128.rs
@@ -13,6 +13,11 @@ pub(crate) fn f16_to_f32(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value
     fx.lib_call("__extendhfsf2", vec![arg_ty], vec![AbiParam::new(types::F32)], &[value])[0]
 }
 
+fn f16_to_f64(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value {
+    let ret = f16_to_f32(fx, value);
+    fx.bcx.ins().fpromote(types::F64, ret)
+}
+
 pub(crate) fn f32_to_f16(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value {
     let ret_ty = if fx.tcx.sess.target.vendor == "apple" && fx.tcx.sess.target.arch == "x86_64" {
         types::I16
@@ -22,6 +27,21 @@ pub(crate) fn f32_to_f16(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value
     let ret = fx.lib_call(
         "__truncsfhf2",
         vec![AbiParam::new(types::F32)],
+        vec![AbiParam::new(ret_ty)],
+        &[value],
+    )[0];
+    if ret_ty == types::I16 { fx.bcx.ins().bitcast(types::F16, MemFlags::new(), ret) } else { ret }
+}
+
+fn f64_to_f16(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value {
+    let ret_ty = if fx.tcx.sess.target.vendor == "apple" && fx.tcx.sess.target.arch == "x86_64" {
+        types::I16
+    } else {
+        types::F16
+    };
+    let ret = fx.lib_call(
+        "__truncdfhf2",
+        vec![AbiParam::new(types::F64)],
         vec![AbiParam::new(ret_ty)],
         &[value],
     )[0];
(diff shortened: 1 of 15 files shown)


src/intrinsics/mod.rs
```diff
@@ -415,12 +490,32 @@ fn codegen_float_intrinsic_call<'tcx>(
...
+            };
             CValue::by_val(ret_val, fx.layout_of(ty))
         }
         _ => {
+            let temp;
+            let name = match name.strip_suffix("f128") {
+                Some(base) if intrinsics::long_double_is_f128(fx.tcx) => {
+                    temp = format!("{base}l");
+                    &temp
+                }
```
tgross35, line 516: I think it may be better to never emit the `*l` functions, only `*f128`. Glibc provides `sinf128`->`sinl` aliases on platforms where that works, so the only downside is linker errors on relevant architectures with musl. libm should fill in the blanks eventually, we do already have a handful of the `*f128` symbols and I'd prefer to avoid adding the `*l` versions.

Cc one of my longest open PRs https://github.com/llvm/llvm-project/pull/76558 to try to fix this in LLVM, and my declined request to musl to provide the same aliases as glibc https://www.openwall.com/lists/musl/2025/03/01/3.

tgross35:
LGTM for the non cranelift-specific parts, LLVM also does promote/demote for f16 powi and fmod.

src/num.rs
```diff
@@ -357,18 +357,32 @@ pub(crate) fn codegen_float_binop<'tcx>(
...
+            let ret_val = fx.lib_call(
                 name,
                 vec![AbiParam::new(ty), AbiParam::new(ty)],
                 vec![AbiParam::new(ty)],
                 &[lhs, rhs],
-            )[0]
+            )[0];
+
+            let ret_val = if *in_lhs.layout().ty.kind() == ty::Float(FloatTy::F16) {
+                fx.bcx.ins().fdemote(types::F16, ret_val)
```
bjorn3, line 381: Wouldn't promote f16 -> f32, fmod, demote f32 -> f16 have rounding issues?

bjorn3:
Is Cranelift ABI compatible with LLVM for f16 and f128?

tgross35:
As a note, once the above lowering to `compiler-builtins` is available, the recently added config at https://github.com/rust-lang/rustc_codegen_cranelift/blob/dfe76361d1cc99b8ce2ee466069b1d1f824dbaa0/src/lib.rs#L200-L208 could likely be updated to start running tests. (Hopefully https://github.com/rust-lang/rust/pull/138087 will merge first to avoid racing on any possible new test failures in r-l/rust).

src/codegen_f16_f128.rs
```diff
@@ -0,0 +1,65 @@
+use crate::prelude::*;
+
+pub(crate) fn f16_to_f32(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value {
+    let (value, arg_ty) =
+        if fx.tcx.sess.target.vendor == "apple" && fx.tcx.sess.target.arch == "x86_64" {
+            (
+                fx.bcx.ins().bitcast(types::I16, MemFlags::new(), value),
+                lib_call_arg_param(fx.tcx, types::I16, false),
+            )
```
bjorn3, line 9: Is this exception part of the apple x86_64 ABI in general? If so maybe moving it into lib_call would be nicer?

src/codegen_f16_f128.rs
```diff
@@ -0,0 +1,65 @@
...
+        } else {
+            (value, AbiParam::new(types::F16))
+        };
+    fx.lib_call("__extendhfsf2", vec![arg_ty], vec![AbiParam::new(types::F32)], &[value])[0]
```
bjorn3, line 13: Can you add these all to the `builtin_functions!` invocation in `compiler_builtins.rs`?

src/codegen_f16_f128.rs
```diff
@@ -117,6 +117,43 @@ pub(crate) fn neg_f128(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value {
...
+    let bits = fx.bcx.ins().band_imm(bits, 0x7fff);
+    fx.bcx.ins().bitcast(types::F16, MemFlags::new(), bits)
+}
+
+pub(crate) fn abs_f128(fx: &mut FunctionCx<'_, '_, '_>, value: Value) -> Value {
+    let bits = fx.bcx.ins().bitcast(types::I128, MemFlags::new(), value);
+    let low = fx.bcx.ins().iconst(types::I64, 0xffff_ffff_ffff_ffff_u64 as i64);
+    let high = fx.bcx.ins().iconst(types::I64, 0x7fff_ffff_ffff_ffff_u64 as i64);
+    let mask = fx.bcx.ins().iconcat(low, high);
+    let bits = fx.bcx.ins().band(bits, mask);
```
bjorn3, line 131: Same here

src/intrinsics/mod.rs
```diff
@@ -400,14 +453,25 @@ fn codegen_float_intrinsic_call<'tcx>(
...
+                sym::floorf16 | sym::floorf32 | sym::floorf64 | sym::floorf128 => {
+                    fx.bcx.ins().floor(args[0])
+                }
+                sym::ceilf16 | sym::ceilf32 | sym::ceilf64 | sym::ceilf128 => {
+                    fx.bcx.ins().ceil(args[0])
+                }
+                sym::truncf16 | sym::truncf32 | sym::truncf64 | sym::truncf128 => {
+                    fx.bcx.ins().trunc(args[0])
+                }
```
bjorn3, line 467: These are already implemented in Cranelift for f16 and f128?

src/codegen_f16_f128.rs
bjorn3: The missing ones are just an oversight.

PR #1579, opened 2025-05-20
Update availability of Cranelift

stefnotch:
Cranelift also ships on Windows these days. This updates the Readme to point that out.

https://rust-lang.github.io/rustup-components-history seems to be an up-to-date source for where Cranelift is available. It seems to be available for both Tier 1 macOS variants, so I changed `x86_64 macOS` to `macOS`

Also, I noticed that the install instructions still say `rustc-codegen-cranelift-preview`, while name name seems to have changed to `rustc-codegen-cranelift`?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -7,7 +7,7 @@ If not please open an issue.
 
 ## Download using Rustup
 
-The Cranelift codegen backend is distributed in nightly builds on Linux and x86_64 macOS. If you want to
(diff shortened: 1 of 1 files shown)


bjorn3:
FWIW the canonical list of targets for which cg_clif is shipped with rustup is: https://github.com/rust-lang/rust/blob/28174fc1a0525888615c8bce8f4ecec13c85ecc6/src/bootstrap/src/utils/helpers.rs#L225-L238

bjorn3:
Thanks!

PR #1580, opened 2025-05-21
Use Cranelift bitcast instead of store & load to bitcast between vectors and non-vectors

beetrees:
The upstream issue https://github.com/bytecodealliance/wasmtime/issues/6104 has been fixed, so the workaround is no longer required.

### DIFF
--- a/src/abi/mod.rs
+++ b/src/abi/mod.rs
@@ -154,10 +154,7 @@ impl<'tcx> FunctionCx<'_, '_, 'tcx> {
 
             let ret = self.lib_call_unadjusted(name, params, returns, &args)[0];
 
-            // FIXME(bytecodealliance/wasmtime#6104) use bitcast instead of store to get from i64x2 to i128
-            let ret_ptr = self.create_stack_slot(16, 16);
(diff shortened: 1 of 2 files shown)


src/abi/mod.rs
```diff
@@ -154,10 +154,7 @@ impl<'tcx> FunctionCx<'_, '_, 'tcx> {
 
             let ret = self.lib_call_unadjusted(name, params, returns, &args)[0];
 
-            // FIXME(bytecodealliance/wasmtime#6104) use bitcast instead of store to get from i64x2 to i128
-            let ret_ptr = self.create_stack_slot(16, 16);
-            ret_ptr.store(self, ret, MemFlags::trusted());
-            Cow::Owned(vec![ret_ptr.load(self, types::I128, MemFlags::trusted())])
+            Cow::Owned(vec![self.bcx.ins().bitcast(types::I128, MemFlags::new(), ret)])
```
bjorn3, line 157: This needs to use codegen_bitcast to ensure that the right endianness flag is set.

bjorn3:
Thanks!

Issue #1528, opened 2024-08-28
Implement -Zembed-source


bjorn3:
https://github.com/rust-lang/rust/pull/126985

PR #1593, opened 2025-09-21
Support -Zembed-source in debuginfo

kehrazy:
Fixes rust-lang/rustc_codegen_cranelift#1528 

- honour the session DWARF version, enabling v5 when embed-source is requested and plumb the flag through the debug context
- attach embedded source bytes to line table FileInfo records alongside existing MD5 hashes
- stop excluding rustc’s embed-source-dwarf run-make test so CG_CLIF exercises the feature

### DIFF
--- a/src/debuginfo/line_info.rs
+++ b/src/debuginfo/line_info.rs
@@ -6,9 +6,7 @@ use std::path::{Component, Path};
 use cranelift_codegen::MachSrcLoc;
 use cranelift_codegen::binemit::CodeOffset;
 use gimli::write::{AttributeValue, FileId, FileInfo, LineProgram, LineString, LineStringTable};
-use rustc_span::{
-    FileName, Pos, SourceFile, SourceFileAndLine, SourceFileHash, SourceFileHashAlgorithm, hygiene,
-};
+use rustc_span::{FileName, Pos, SourceFile, SourceFileAndLine, SourceFileHashAlgorithm, hygiene};
 
 use crate::debuginfo::FunctionDebugContext;
 use crate::debuginfo::emit::address_for_func;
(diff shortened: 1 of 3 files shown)


src/debuginfo/mod.rs
```diff
@@ -67,21 +68,27 @@ impl DebugContext {
             return None;
         }
 
+        // FIXME this should be configurable
+        // macOS doesn't seem to support DWARF > 3
```
bjorn3, line 72:
```suggestion
```

src/debuginfo/line_info.rs
```diff
@@ -44,21 +42,27 @@ fn osstr_as_utf8_bytes(path: &OsStr) -> &[u8] {
...
-        Some(FileInfo {
-            timestamp: 0,
-            size: 0,
-            md5: buf,
-            source: None, // FIXME implement -Zembed-source
-        })
-    } else {
-        None
+fn make_file_info(source_file: &SourceFile, embed_source: bool) -> Option<FileInfo> {
+    let has_md5 = source_file.src_hash.kind == SourceFileHashAlgorithm::Md5;
```
bjorn3, line 46: Do we actually need the MD5 hash when using `-Zembed-source`?

bjorn3:
Thanks for working on this!

kehrazy:
> Unrelated to this PR, but while trying to test this PR I noticed that with DWARFv5 anonymous sources get invalid file names. They are supposed to be called `<anon>`, but get called `/path/to/working_directory/<anon>` instead by gdb.

The fix might be [FileName::prefer_remapped_unconditionally](https://doc.rust-lang.org/beta/nightly-rustc/src/rustc_span/lib.rs.html#423-425) - internal files and all that.

How do you check the output? With what `RUSTFLAGS` do you compile a binary? Can you give me a repro?

bjorn3:
I used `dist/rustc-clif -g -Zdwarf-version=5 -Zembed-source foo.rs` and then removing `foo.rs` for trying to test `-Zembed-source`.

`echo 'fn main() {}' | dist/rustc-clif -g -Zdwarf-version=5 -` reproduces the `<anon>` issue.

kehrazy:
> which looks okay!

I really hate this solution, though, and I think it should be reverted - I can't seem to find a better approach here.

PR #1634, opened 2026-03-12
Fix __eh_frame for AArch64 on macOS

philipc:
The relocations must use symbols (not sections), and the relative relocation needs ARM64_RELOC_SUBTRACTOR.

The section must be in __TEXT rather than __DWARF, which is handled by using StandardSection::EhFrame. This also changes the section type for ELF x86-64.

Depends on https://github.com/gimli-rs/object/pull/856

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,27 +43,27 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "cranelift-assembler-x64"
-version = "0.130.0"
+version = "0.131.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4f248321c6a7d4de5dcf2939368e96a397ad3f53b6a076e38d0104d1da326d37"
+checksum = "6edb5bdd1af46714e3224a017fabbbd57f70df4e840eb5ad6a7429dc456119d6"
 dependencies = [
  "cranelift-assembler-x64-meta",
 ]
 
 [[package]]
 name = "cranelift-assembler-x64-meta"
-version = "0.130.0"
+version = "0.131.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ab6d78ff1f7d9bf8b7e1afbedbf78ba49e38e9da479d4c8a2db094e22f64e2bc"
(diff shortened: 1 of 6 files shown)


Cargo.lock
```diff
@@ -277,6 +277,15 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "hashbrown"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4f467dd6dccf739c208452f8014c75c18bb8301b050ad1cfb27153803edb0f51"
+dependencies = [
+ "foldhash",
+]
```
bjorn3, line 287: Is it possible to update other crates to avoid two versions of hashbrown?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1634/720aed373a6e31583c10681384a06143c5d64a95..7fddef64f1c328d804736a39602d96d7f74cb168)*

src/debuginfo/emit.rs
```diff
@@ -43,7 +43,7 @@ impl DebugContext {
         let _: Result<()> = sections.for_each(|id, section| {
             if let Some(section_id) = section_map.get(&id) {
                 for reloc in &section.relocs {
-                    product.add_debug_reloc(&section_map, section_id, reloc);
```
bjorn3, line 46: This should be true, right? LLVM emits section relocations on ELF at least.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1634/720aed373a6e31583c10681384a06143c5d64a95..7fddef64f1c328d804736a39602d96d7f74cb168)*

bjorn3:
CI failure is because of https://github.com/rust-lang/cargo/issues/16925

src/debuginfo/emit.rs
philipc: Good catch, I didn't intend to change the behaviour here.

Cargo.lock
philipc:
```
hashbrown v0.16.1
├── cranelift-codegen v0.131.0
├── gimli v0.33.0
└── wasmtime-internal-core v44.0.0
```
wasmtime and gimli both have merged changes to update hashbrown, but need new releases.

bjorn3 (approved):

PR #1637, opened 2026-03-26
toolchain: avoid broad substring replacement in get_toolchain_binary

### LINKED ISSUE #1636: `get_toolchain_binary` can derive the wrong tool name by doing plain substring replacement
`src/toolchain.rs` derives tool names from the linker filename by doing plain `.replace("ld", tool).replace("gcc", tool).replace("clang", tool).replace("cc", tool)` on the full filename.

This is a pretty loose transformation and can rewrite unintended substrings instead of just replacing the compiler/linker suffix. For example, it assumes the tool name can always be inferred by global text replacement inside the linker basename.

That makes the helper fragile for less standard linker names / wrappers.

Expected behavior: only rewrite known linker basename patterns, instead of replacing arbitrary substrings in the whole filename.

  Code: `src/toolchain.rs`

AayushMainali-Github:
Fixes rust-lang/rustc_codegen_cranelift#1636
  
  ## Summary

  Make `get_toolchain_binary()` only rewrite recognized linker suffixes instead of doing broad substring replacement on the whole linker filename.

  ## Details

  The old code chained `.replace("ld", ...)`, `.replace("gcc", ...)`, `.replace("clang", ...)`, and `.replace("cc", ...)`, which could rewrite unrelated parts of the linker name.

  This changes the logic to only handle known endings:

  - `gcc`
  - `clang`
  - `cc`
  - `ld`

  The existing `ld.lld` special case is kept.

### DIFF
--- a/src/toolchain.rs
+++ b/src/toolchain.rs
@@ -11,18 +11,22 @@ pub(crate) fn get_toolchain_binary(sess: &Session, tool: &str) -> PathBuf {
     let linker_file_name =
         linker.file_name().unwrap().to_str().expect("linker filename should be valid UTF-8");
 
-    if linker_file_name == "ld.lld" {
-        if tool != "ld" {
-            linker.set_file_name(tool)
-        }
+    let tool_file_name = if linker_file_name == "ld.lld" {
+        if tool == "ld" { "ld.lld".to_owned() } else { tool.to_owned() }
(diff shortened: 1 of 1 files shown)


bjorn3:
As explained in https://github.com/rust-lang/rustc_codegen_cranelift/issues/1636#issuecomment-4137454337 I don't think this is worth it as it only gets used when `CG_CLIF_FORCE_GNU_AS` is set.

PR #1638, opened 2026-04-10
Fix compilation for custom JSON targets

Finchiedev:
A very simple change to allow compilation with custom JSON targets. Previously the Cranelift backend (and not LLVM) would produce an error as the `rustc` assembler process expected `-Zunstable-options` when JSON targets were used.

Skipped doing fancier logic as always passing `-Zunstable-options` seems to work fine, but could be wrong! Happy to change things if I'm wrong about how that works :)

### DIFF
--- a/src/global_asm.rs
+++ b/src/global_asm.rs
@@ -235,6 +235,9 @@ pub(crate) fn compile_global_asm(
             .arg("-")
             .arg("-Abad_asm_style")
             .arg("-Zcodegen-backend=llvm")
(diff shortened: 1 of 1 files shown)


bjorn3 (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1638/adb7c3f538e0510a6cc434a2883d7679b014d891..70ee124a0b3f2230a3676c5b54c295b5220e9970)*

Issue #1465, opened 2024-03-16
Copy byval argument to local stackslot if alignment is insufficient


bjorn3:
See https://github.com/rust-lang/rust/pull/122212

PR #1641, opened 2026-04-13
Copy byval argument to local stackslot if alignment is insufficient

0xmuon:
Copy the underaligned `byval` (indirect) arguments into a local stackslot when the incoming pointer alignment is less than the Rust ABI alignment. This avoids miscompiles from assuming stronger alignment than the ABI guarantees.

### DIFF
--- a/src/abi/mod.rs
+++ b/src/abi/mod.rs
@@ -12,7 +12,7 @@ use cranelift_codegen::ir::{
 };
 use cranelift_codegen::isa::CallConv;
 use cranelift_module::ModuleError;
-use rustc_abi::{CanonAbi, ExternAbi, X86Call};
+use rustc_abi::{Align, CanonAbi, ExternAbi, X86Call};
 use rustc_codegen_ssa::base::is_call_from_compiler_builtins_to_upstream_monomorphization;
 use rustc_codegen_ssa::errors::CompilerBuiltinsCannotCall;
 use rustc_middle::middle::codegen_fn_attrs::CodegenFnAttrFlags;
@@ -243,10 +243,17 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
         self::returning::codegen_return_param(fx, &ssa_analyzed, &mut block_params_iter);
     assert_eq!(fx.local_map.push(ret_place), RETURN_PLACE);
 
-    // None means pass_mode == NoPass
+    struct ArgValue<'tcx> {
+        value: Option<CValue<'tcx>>,
+        /// If set, the argument is a byval/byref pointer whose pointee alignment is smaller than
+        /// the Rust ABI alignment. In that case the argument must be copied to a sufficiently
+        /// aligned local stack slot before it can be treated as a place.
+        underaligned_pointee_align: Option<Align>,
+    }
+
     enum ArgKind<'tcx> {
-        Normal(Option<CValue<'tcx>>),
-        Spread(Vec<Option<CValue<'tcx>>>),
+        Normal(ArgValue<'tcx>),
+        Spread(Vec<ArgValue<'tcx>>),
     }
 
     // FIXME implement variadics in cranelift
@@ -280,17 +287,34 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
                 let mut params = Vec::new();
                 for (i, _arg_ty) in tupled_arg_tys.iter().enumerate() {
                     let arg_abi = arg_abis_iter.next().unwrap();
-                    let param =
+                    let value =
                         cvalue_for_param(fx, Some(local), Some(i), arg_abi, &mut block_params_iter);
-                    params.push(param);
+                    let underaligned_pointee_align =
+                        match arg_abi.mode {
+                            PassMode::Indirect { attrs, .. } => attrs
+                                .pointee_align
+                                .filter(|&pointee_align| pointee_align < arg_abi.layout.align.abi),
+                            _ => None,
+                        };
+                    params.push(ArgValue { value, underaligned_pointee_align });
                 }
 
                 (local, ArgKind::Spread(params), arg_ty)
             } else {
                 let arg_abi = arg_abis_iter.next().unwrap();
-                let param =
+                let value =
                     cvalue_for_param(fx, Some(local), None, arg_abi, &mut block_params_iter);
-                (local, ArgKind::Normal(param), arg_ty)
+                let underaligned_pointee_align = match arg_abi.mode {
+                    PassMode::Indirect { attrs, .. } => attrs
+                        .pointee_align
+                        .filter(|&pointee_align| pointee_align < arg_abi.layout.align.abi),
+                    _ => None,
+                };
+                (
+                    local,
+                    ArgKind::Normal(ArgValue { value, underaligned_pointee_align }),
+                    arg_ty,
+                )
             }
         })
         .collect::<Vec<(Local, ArgKind<'tcx>, Ty<'tcx>)>>();
@@ -311,7 +335,9 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
     for (local, arg_kind, ty) in func_params {
         // While this is normally an optimization to prevent an unnecessary copy when an argument is
         // not mutated by the current function, this is necessary to support unsized arguments.
-        if let ArgKind::Normal(Some(val)) = arg_kind {
+        if let ArgKind::Normal(ArgValue { value: Some(val), underaligned_pointee_align: None }) =
+            arg_kind
+        {
             if let Some((addr, meta)) = val.try_to_ptr() {
                 // Ownership of the value at the backing storage for an argument is passed to the
                 // callee per the ABI, so it is fine to borrow the backing storage of this argument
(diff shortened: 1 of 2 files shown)


src/abi/mod.rs
```diff
@@ -280,17 +287,29 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
                 let mut params = Vec::new();
                 for (i, _arg_ty) in tupled_arg_tys.iter().enumerate() {
                     let arg_abi = arg_abis_iter.next().unwrap();
-                    let param =
+                    let value =
                         cvalue_for_param(fx, Some(local), Some(i), arg_abi, &mut block_params_iter);
```
bjorn3, line 289: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1641/86354453e582bbffb261ae9e6d6d7d7775cba9c7..c153f587cd69316058aa62a93ae9802cbe921b0f)*

Maybe return `ArgValue` from `cvalue_for_param` instead?

src/abi/mod.rs
```diff
@@ -336,15 +357,74 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
...
+                        to_addr,
+                        from_addr,
+                        size,
+                        dst_align,
+                        src_align,
+                        true,
+                        flags,
+                    );
+                } else {
                     place.write_cvalue(fx, param);
```
bjorn3: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1641/86354453e582bbffb261ae9e6d6d7d7775cba9c7..c153f587cd69316058aa62a93ae9802cbe921b0f)*

Hang on, doesn't this `write_cvalue` already copy to a stack slot of sufficient alignment? I think really the only thing necessary is having `if let ArgKind::Normal(Some(val)) = arg_kind` only apply when the argument is correctly aligned. Or to handle alignment correctly, when the alignment of the param doesn't match the alignment of the type, use a type of `[u8; size_of::<T>]` for `param` in `cvalue_for_param` and use `write_cvalue_maybe_transmute` to write it to `place` in that case. That would also ensure the case where the ABI passes it byval while inside the function it is stored as SSA value rather than in a stack slot.

bjorn3 (changes requested):

src/abi/mod.rs
0xmuon: Do you agree this covers the SSA-local case correctly, and should we also handle `meta_attrs: Some(_)` (unsized indirect) or leave that unchanged for now?

0xmuon: so, in `abi::codegen_fn_prelude`, the `if let ArgKind::Normal(Some(val))` fast-path to be gated on `underaligned_pointee_align.is_none()` (like only when `attrs.pointee_align >= layout.align.abi` or `pointee_align` is `None`) and in `abi::pass_mode::cvalue_for_param`, for `PassMode::Indirect { attrs, meta_attrs: None, ......}` where `attrs.pointee_align < arg_abi.layout.align.abi`, we return `CValue::by_ref(ptr, layout_of([u8; size]))` instead of `layout_of(T)` and when initializing locals from such params we use `place.write_cvalue_transmute(fx, param)` (and similarly for spread fields). Am i right'?

src/abi/pass_mode.rs
```diff
@@ -318,9 +320,22 @@ pub(super) fn cvalue_for_param<'tcx>(
...
-            Some(CValue::by_ref(Pointer::new(block_params[0]), arg_abi.layout))
+            if let Some(pointee_align) = attrs.pointee_align
+                && pointee_align < arg_abi.layout.align.abi
+                && arg_abi.layout.is_sized()
+                && arg_abi.layout.size != Size::ZERO
+            {
+                underaligned_pointee_align = Some(pointee_align);
+                // Underaligned pointer: treat as `[u8; size]` and transmute-copy into the real type.
+                let len = ty::Const::from_target_usize(fx.tcx, arg_abi.layout.size.bytes());
+                let bytes_ty = fx.tcx.mk_ty_from_kind(ty::Array(fx.tcx.types.u8, len));
```
bjorn3: This can use `Ty::new_array()`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1641/86354453e582bbffb261ae9e6d6d7d7775cba9c7..067bf06ed19043fe91f563ef9c51ddfc8d114e14)*

src/abi/mod.rs
```diff
@@ -31,6 +31,11 @@ use crate::base::codegen_unwind_terminate;
 use crate::debuginfo::EXCEPTION_HANDLER_CLEANUP;
 use crate::prelude::*;
 
+pub(super) struct ArgValue<'tcx> {
+    pub(super) value: Option<CValue<'tcx>>,
+    pub(super) underaligned_pointee_align: Option<Align>,
```
bjorn3: This can now be a bool `is_underaligned_pointee`.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1641/86354453e582bbffb261ae9e6d6d7d7775cba9c7..067bf06ed19043fe91f563ef9c51ddfc8d114e14)*

src/abi/mod.rs
```diff
@@ -243,10 +248,9 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
         self::returning::codegen_return_param(fx, &ssa_analyzed, &mut block_params_iter);
     assert_eq!(fx.local_map.push(ret_place), RETURN_PLACE);
 
-    // None means pass_mode == NoPass
     enum ArgKind<'tcx> {
-        Normal(Option<CValue<'tcx>>),
-        Spread(Vec<Option<CValue<'tcx>>>),
+        Normal(ArgValue<'tcx>),
+        Spread(Vec<ArgValue<'tcx>>),
```
bjorn3: Maybe pull out the `Option` around the `field` ty and have:
```suggestion
        Normal(Option<ArgValue<'tcx>>),
        Spread(Vec<Option<ArgValue<'tcx>>>),
```
? The align field is unused when the argument is not passed at all. I don't know how much that would complicate `cvalue_for_param` though.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1641/86354453e582bbffb261ae9e6d6d7d7775cba9c7..067bf06ed19043fe91f563ef9c51ddfc8d114e14)*

src/abi/mod.rs
```diff
@@ -299,8 +306,8 @@ pub(crate) fn codegen_fn_prelude<'tcx>(fx: &mut FunctionCx<'_, '_, 'tcx>, start_
     if fx.instance.def.requires_caller_location(fx.tcx) {
         // Store caller location for `#[track_caller]`.
         let arg_abi = arg_abis_iter.next().unwrap();
-        fx.caller_location =
-            Some(cvalue_for_param(fx, None, None, arg_abi, &mut block_params_iter).unwrap());
+        let arg = cvalue_for_param(fx, None, None, arg_abi, &mut block_params_iter);
```
bjorn3: Maybe assert that `arg` is not underaligned?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1641/86354453e582bbffb261ae9e6d6d7d7775cba9c7..067bf06ed19043fe91f563ef9c51ddfc8d114e14)*

bjorn3 (changes requested):

src/abi/mod.rs
bjorn3: Yes

src/abi/mod.rs
0xmuon: I considered that, but since `cvalue_for_param` now returns a single `ArgValue`, keeping `Option` outside would make the call sites and caller-location path a bit more awkward again. Since the underaligned flag is now just a `bool`, I kept `ArgKind::{Normal,Spread}` holding `ArgValue` directly and use `value: None` for `PassMode::Ignore`.

src/abi/mod.rs
0xmuon: Yep.I will add an assertion there before unwrapping the value.

src/abi/mod.rs
0xmuon: @bjorn3 I only need to know whether to avoid the fast-path or use the transmute-write path, so that I will change this to a bool `is_underaligned_pointee`.

0xmuon: @bjorn3 could you please guide...

bjorn3: What is the problem you want help with?

0xmuon: in [src/abi/pass_mode.rs](https://github.com/0xmuon/rustc_codegen_cranelift/blob/main/src/abi/pass_mode.rs) I still construct the byte-array type with `mk_ty_from_kind(ty::Array(..))`, and you suggested using `Ty::new_array()`. I am unsure what the preferred pattern is here (and whether there is any gotcha with `ty::Const::from_target_usize` or `layout_of`), so I wanted guidance before pushing the final fix.

bjorn3: `mk_ty_from_kind` works, but `Ty::new_array` is simpler. You can pass it the length as a regular u64.

src/abi/mod.rs
bjorn3: I pushed a commit for this. I think it is a bit clearer.

bjorn3:
I just found out that this doesn't actually fix #1465. It fixes a similar problem, but on the other side of the ABI. This fixes a problem on the callee when the ABI uses a lower alignment than the type needs, while #1465 is about copying on the caller side when the ABI requires a higher alignment than the type. The only tests in https://github.com/rust-lang/rust/pull/122212 for the callee side bug only works on 32bit x86 which cg_clif doesn't support, so I can't actually test if your PR properly fixes the ABI mismatch. I do think it is correct, so I will merge it once CI passes. Thanks for working on this!

PR #1644, opened 2026-04-14
fail loudly on unsized by-ref invariants

0xmuon:
replace the remaining `todo!()` paths for unsized by-ref values in `src/value_and_place.rs` with explicit `bug!()` failures.
This makes invariant violations fail loudly instead of panicking via unfinished code paths,and without claiming support for unsized by-ref handling that is not actually implemented.

### DIFF
--- a/src/value_and_place.rs
+++ b/src/value_and_place.rs
@@ -204,7 +204,14 @@ impl<'tcx> CValue<'tcx> {
                 let (field_ptr, field_layout) = codegen_field(fx, ptr, None, layout, field);
                 CValue::by_ref(field_ptr, field_layout)
             }
-            CValueInner::ByRef(_, Some(_)) => todo!(),
+            CValueInner::ByRef(ptr, Some(extra)) => {
+                let (field_ptr, field_layout) = codegen_field(fx, ptr, Some(extra), layout, field);
+                if fx.tcx.type_has_metadata(field_layout.ty, ty::TypingEnv::fully_monomorphized()) {
+                    CValue::by_ref_unsized(field_ptr, extra, field_layout)
+                } else {
+                    CValue::by_ref(field_ptr, field_layout)
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/value_and_place.rs
```diff
@@ -655,7 +662,26 @@ impl<'tcx> CPlace<'tcx> {
                             flags,
                         );
                     }
-                    CValueInner::ByRef(_, Some(_)) => todo!(),
+                    CValueInner::ByRef(from_ptr, Some(_extra)) => {
+                        // Unsized values shouldn't normally be written into sized places. However,
+                        // if this happens, we can still copy the sized prefix using the destination layout's fixed size.
```
bjorn3: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1644/86354453e582bbffb261ae9e6d6d7d7775cba9c7..7a8f3afaa5de33324d0171d55d065bdfc050297f)*

If an IR invariant is violated, we should fail loudly, not silently do something that seems roughly correct, but almost certainly causes hard to debug corruption.

0xmuon:
Is “copy sized prefix when writing from `ByRef(_, Some(meta))` into a sized place” fallback acceptable or would you prefer we `span_bug!` there instead to enforce that such writes never happen?

src/value_and_place.rs
```diff
@@ -204,7 +204,14 @@ impl<'tcx> CValue<'tcx> {
...
             }
-            CValueInner::ByRef(_, Some(_)) => todo!(),
+            CValueInner::ByRef(ptr, Some(extra)) => {
+                let (field_ptr, field_layout) = codegen_field(fx, ptr, Some(extra), layout, field);
+                if fx.tcx.type_has_metadata(field_layout.ty, ty::TypingEnv::fully_monomorphized()) {
+                    CValue::by_ref_unsized(field_ptr, extra, field_layout)
+                } else {
+                    CValue::by_ref(field_ptr, field_layout)
+                }
+            }
```
bjorn3, line 209: *[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1644/86354453e582bbffb261ae9e6d6d7d7775cba9c7..7a8f3afaa5de33324d0171d55d065bdfc050297f)*

The only places where `value_field` is used don't use unsized values afaik. It can always be implemented later if necessary.

src/value_and_place.rs
0xmuon: yup agreed! will correct

bjorn3 (approved):
Please update the PR title and description to match the contents of the PR. Other than that LGTM.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1644/86354453e582bbffb261ae9e6d6d7d7775cba9c7..2237da26521fe1fec63c373b729ac41c69b1d02b)*

0xmuon:
Done!

PR #1648, opened 2026-04-19
Parse and set Mach-O section type and attributes

### LINKED ISSUE #1588: add support for mod_init_funcs (was compat: rust-ctor fails to trigger with Cranelift on MacOS, but works with LLVM)
I noticed when switching to the cranelift backend that my `typetag` types stopped working:
https://github.com/dtolnay/typetag/issues/35

I'm not sure what exactly is to blame, but figured it warranted reporting as a compat issue. Through some subsequent debugging, I've identified noticed an issue specific to Cranelift on MacOS where the `rust-ctor` crate to fails to trigger `#[ctor]`-annotated code on module initialization: https://github.com/mmastrac/rust-ctor/issues/339

Other thoughts:

- There was some discussion in https://github.com/rust-lang/rust/issues/99721 that it could be due to issues with rustc and static linking, but given this is cranelift-specific (switching to LLVM causes typetag and rust-ctor to work as expected) I'm skeptical.
- The `inventory` crate uses a technique similar to ctor so I bet https://github.com/rust-lang/rustc_codegen_cranelift/issues/1577 is also bit by the same root cause affecting this issue.

### LINKED ISSUE #20785: macOS: main branch fails to compile with cranelift (unsupported section type mod_init_funcs)
```
error: #[link_section = "__mod_init_func"] is not supported: unsupported section type mod_init_funcs

error: could not compile `bevy_reflect` (lib) due to 1 previous error
```

## Bevy version and features

- main branch commit `cbf989c`

## Relevant system information

running with `cargo 1.91.0-nightly (a6c58d430 2025-08-26)` on macOS 15.6

## Steps to reproduce

create an empty project with bevy set to track the main branch and add cranelift to your `.cargo/config.toml`
```toml
[unstable]
codegen-backend = true

[profile.dev]
codegen-backend = "cranelift"
```

## What went wrong

Project fails to compile `bevy_reflect` with the error
```
error: #[link_section = "__mod_init_func"] is not supported: unsupported section type mod_init_funcs

error: could not compile `bevy_reflect` (lib) due to 1 previous error
```

## Related issues

This appears to be an issue in cranelift itself 
- https://github.com/rust-lang/rustc_codegen_cranelift/issues/1588

That being said (i think) downstream crates found workarounds? 
- https://github.com/dtolnay/inventory/issues/82

im not very knowledgeable on either of these topics however so i cant say for sure.

madsmtm:
Mach-O sections have a 32-bit "flags" parameter, which contains 8 bits for a "section type", and 24 bits for "attributes". These can be controlled with `#[link_section = ...]` like so:
```rust
// Sets the section type to S_MOD_INIT_FUNC_POINTERS
#[unsafe(link_section = "__DATA,__mod_init_func,mod_init_funcs")]
static CONSTRUCTOR: extern "C" fn() = ...;

// Sets the section type to S_CSTRING_LITERALS
#[link_section = "__TEXT,__cstring,cstring_literals"]
static NAME_DATA: [u8; 4] = b"foo\0";

#[unsafe(link_section = "__TEXT,__text,regular,pure_instructions+no_dead_strip")]
static LINKER_USED_INSTRUCTIONS: [u8; 10] = ...;
```

The default section type is "regular" (`S_REGULAR`).

This PR parses the section type names and attribute names understood by LLVM, and convert them to the corresponding Mach-O flags.

Fixes https://github.com/rust-lang/rustc_codegen_cranelift/issues/1588.
Fixes https://github.com/bevyengine/bevy/issues/20785.
Depends on https://github.com/bytecodealliance/wasmtime/pull/13137.

Related to https://github.com/rust-lang/rust/pull/155065, we could also put this parsing logic higher up, to rely less on LLVM / get diagnostics with spans for this on all backends?

### DIFF
--- a/src/constant.rs
+++ b/src/constant.rs
@@ -414,49 +414,7 @@ fn define_all_allocs(tcx: TyCtxt<'_>, module: &mut dyn Module, cx: &mut Constant
         data.set_align(alloc.align.bytes());
 
         if let Some(section_name) = section_name {
-            let (segment_name, section_name) = if tcx.sess.target.is_like_darwin {
-                // See https://github.com/llvm/llvm-project/blob/main/llvm/lib/MC/MCSectionMachO.cpp
-                let mut parts = section_name.as_str().split(',');
-                let Some(segment_name) = parts.next() else {
-                    tcx.dcx().fatal(format!(
-                        "#[link_section = \"{}\"] is not valid for macos target: must be segment and section separated by comma",
-                        section_name
-                    ));
-                };
-                let Some(section_name) = parts.next() else {
-                    tcx.dcx().fatal(format!(
-                        "#[link_section = \"{}\"] is not valid for macos target: must be segment and section separated by comma",
-                        section_name
-                    ));
-                };
-                if section_name.len() > 16 {
-                    tcx.dcx().fatal(format!(
-                        "#[link_section = \"{}\"] is not valid for macos target: section name bigger than 16 bytes",
-                        section_name
-                    ));
-                }
-                let section_type = parts.next().unwrap_or("regular");
-                if section_type != "regular" && section_type != "cstring_literals" {
-                    tcx.dcx().fatal(format!(
-                        "#[link_section = \"{}\"] is not supported: unsupported section type {}",
-                        section_name, section_type,
-                    ));
-                }
-                let _attrs = parts.next();
-                if parts.next().is_some() {
-                    tcx.dcx().fatal(format!(
-                        "#[link_section = \"{}\"] is not valid for macos target: too many components",
-                        section_name
-                    ));
-                }
-                // FIXME(bytecodealliance/wasmtime#8901) set S_CSTRING_LITERALS section type when
-                // cstring_literals is specified
-                (segment_name, section_name)
-            } else {
-                ("", section_name.as_str())
-            };
-            // FIXME pass correct section flags on Mach-O
-            data.set_segment_section(segment_name, section_name, 0);
+            data.set_custom_section(section_name.as_str());
         }
 
         let bytes = alloc.inspect_with_uninit_and_ptr_outside_interpreter(0..alloc.len()).to_vec();
--- a/src/debuginfo/unwind.rs
+++ b/src/debuginfo/unwind.rs
@@ -204,7 +204,7 @@ impl UnwindContext {
 
                     let mut data = DataDescription::new();
                     data.define(gcc_except_table.writer.into_vec().into_boxed_slice());
-                    data.set_segment_section("", ".gcc_except_table", 0);
+                    data.set_custom_section(".gcc_except_table");
 
                     for reloc in &gcc_except_table.relocs {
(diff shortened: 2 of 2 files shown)


src/constant.rs
```diff
@@ -649,3 +661,45 @@ pub(crate) fn mir_operand_get_const_val<'tcx>(
         }
     }
 }
+
+// We support the same custom section type / attrs naming as LLVM:
+// <https://github.com/llvm/llvm-project/blob/llvmorg-22.1.3/llvm/lib/MC/MCSectionMachO.cpp#L23-L91>
+// <https://github.com/llvm/llvm-project/blob/llvmorg-22.1.3/llvm/include/llvm/BinaryFormat/MachO.h#L120-L223>
+const MACHO_SECTION_TYPES: &[(&str, u32)] = &[
```
madsmtm: I did this as a table originally because I kinda wanted to print the available options / have suggestions for mistyped section types, but I ended up not implementing that, because I didn't really know if there were a good way to do so with `rustc`'s diagnostic APIs?

I can do a `match` instead if you prefer?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1648/973ab0a729022a5156588f2064099d855283c4a9..27835598796ad0cb56a3b817785a9945739ccca9)*

bjorn3: Using tables here is fine by me.

src/constant.rs
```diff
@@ -649,3 +661,45 @@ pub(crate) fn mir_operand_get_const_val<'tcx>(
...
+    ("regular", object::macho::S_REGULAR),
```
bjorn3: Maybe `use object::macho`? Or even put those two tables in an inline submodule with `use object::macho::*`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1648/9446d0db62449d6c0cc0ba56f33b45b564e9c35e..27835598796ad0cb56a3b817785a9945739ccca9)*

bjorn3:
I updated Cranelift earlier today, so it should work after a rebase.

bjorn3:
@madsmtm did you intend to work on this further or do you want me to pick it up?

madsmtm:
I've opened https://github.com/bytecodealliance/wasmtime/pull/13869 instead, since on further reflection I think it makes more sense to have the parsing in `cranelift-object` (it's useful for other consumers of `cranelift-object`). I've updated this PR to what it'd look like after that.

bjorn3:
Looks like this change will have to be merged at the same time as updating Cranelift. I've added it to the update_cranelift branch and added you as co-author. The branch will be merged in about two weeks when the next Cranelift version is released.

madsmtm:
Coolio!

bjorn3:
Included in https://github.com/rust-lang/rustc_codegen_cranelift/commit/2bc05d45078dbee6495c008cbd5d1dc24edff53f.

PR #1666, opened 2026-06-11
Implement remaining aarch64 SHA-3 LLVM intrinsics

CathalMullan:
Implements the remaining SHA-3 intrinsics.

Needed for ML-KEM / ML-DSA support. 
~~`graviola` has a feature flag which makes use of `libcrux` for PQC. Those tests aren't currently run, but can be enabled in the future (once a few more missing intrinsics are added).~~ - No longer the case, `graviola` switched to a native ML-KEM implementation.

### DIFF
--- a/src/intrinsics/llvm_aarch64.rs
+++ b/src/intrinsics/llvm_aarch64.rs
@@ -977,6 +977,88 @@ pub(super) fn codegen_aarch64_llvm_intrinsic_call<'tcx>(
             );
         }
 
+        "llvm.aarch64.crypto.eor3s.v2i64"
+        | "llvm.aarch64.crypto.eor3s.v4i32"
+        | "llvm.aarch64.crypto.eor3s.v8i16"
+        | "llvm.aarch64.crypto.eor3s.v16i8"
+        | "llvm.aarch64.crypto.eor3u.v2i64"
+        | "llvm.aarch64.crypto.eor3u.v4i32"
+        | "llvm.aarch64.crypto.eor3u.v8i16"
+        | "llvm.aarch64.crypto.eor3u.v16i8" => {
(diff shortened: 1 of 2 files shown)


src/intrinsics/llvm_aarch64.rs
```diff
@@ -977,6 +977,88 @@ pub(super) fn codegen_aarch64_llvm_intrinsic_call<'tcx>(
...
+
+            simd_trio_for_each_lane(
+                fx,
+                a,
+                b,
+                c,
+                ret,
+                &|fx, _lane_ty, _res_lane_ty, a_lane, b_lane, c_lane| {
+                    let band_not = fx.bcx.ins().band_not(b_lane, c_lane);
+                    fx.bcx.ins().bxor(a_lane, band_not)
```
bjorn3, line 1023: Isn't this supposed to be `b ^ (a & !c)` rather than `a ^ (b & !c)`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1666/7227b9ca90799dda0b1501120323b9b392bffcf8..2c81a777737ca607f199d04fd7509c5df15f9147)*

CathalMullan: I think the current order is right. Testing against `stdarch`, `b ^ (a & !c)` results in 8 failing `vbcaxq` tests, while the current approach passes all.

bjorn3: Ah, the instruction is `bcax d, n, m, a`, but the operation in the docs lists the operands as m, n, a, which indeed results in `a ^ (b & !c)` as correct order.

bjorn3 (approved):

PR #1667, opened 2026-06-14
Integrate `stdarch` tests for aarch64

CathalMullan:
Had a go at wiring up `stdarch`. Runs the `core_arch` and `intrinsic-test` test suites. 

Looked into x86 as well. It had more hard errors to resolve up front, so opted to stick with aarch64 for now.

The skip mechanism for these may be a little cumbersome to maintain.
Maybe making the tests opt-in rather than opt-out would be the better approach?


build_system/tests.rs
```diff
@@ -162,6 +162,19 @@ static SYSROOT_TESTS_SRC: RelPath = RelPath::build("sysroot_tests");
...
+pub(crate) static STDARCH_REPO: GitRepo = GitRepo::github(
+    "rust-lang",
+    "stdarch",
+    "afc40bbc4fde7999324bb58fba692b6a51dc9290",
+    &[],
+    "62c59d72e1e4b348",
+    "stdarch",
+);
+
+static STDARCH: CargoProject = CargoProject::new(STDARCH_REPO.source_dir(), "stdarch_target");
```
bjorn3: It should be possible to use the version from the standard library sources just like for portable-simd.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1667/7227b9ca90799dda0b1501120323b9b392bffcf8..a4adef342a9b88963b2f7837d2178de4f03d2ea1)*

CathalMullan: Looks like the stdarch intrinsic-test crate gets stripped from the source before bundling: https://github.com/rust-lang/rust/blob/main/src/bootstrap/src/core/build_steps/dist.rs#L1204-L1208

bjorn3: Right, I added those exclusions myself :facepalm:

bjorn3:
> Maybe making the tests opt-in rather than opt-out would be the better approach?

Yeah, I think that would work better. Aside from currently being a shorter list, it also avoids mistakenly thinking a new failing test got added that can be disabled just fine rather than an existing test now failing.

build_system/tests.rs
CathalMullan: Turns out this is outdated anyways, the heavy intrinsics data now lives in `library/stdarch/intrinsics_data`, which isn't stripped. Also looks to be smaller than the 30MB it was in the past, around 12MB now, compresses to 300KB.

Would be simpler to remove the filter from upstream.

CathalMullan:
Now uses `rust-src` for the `stdarch` tests.

Switching from opt-out to opt-in will require some upstream work in `stdarch`. It only supports `STDARCH_TEST_SKIP_FUNCTION` / `--skip`, no mechanism to only run a subset of tests.

The opt-in lists would likely be longer than the existing skip lists fwiw.

build_system/tests.rs
```diff
@@ -262,6 +268,118 @@ const EXTENDED_SYSROOT_SUITE: &[TestCase] = &[
...
+            eprintln!("Cross-Compiling: Not running tests");
+            return;
+        }
+
+        let cc = if triple.contains("apple") { "clang" } else { "gcc" };
+
+        let mut generate_cmd = STDARCH.run(&runner.target_compiler, &runner.dirs);
+        generate_cmd.current_dir(STDARCH.source_dir(&runner.dirs));
+        generate_cmd.args([
+            "--release",
```
bjorn3: Is the overall test time more or less when removing `--release`?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1667/c232e6289b7757783926a9970618814948f9457d..180f2122291155a0d90f0293acf33c4c2a448b22)*

build_system/tests.rs
```diff
@@ -262,6 +268,118 @@ const EXTENDED_SYSROOT_SUITE: &[TestCase] = &[
...
+        if skip.exists() {
+            generate_cmd.arg("--skip").arg(skip);
+        }
+
+        spawn_and_wait(generate_cmd);
+
+        let mut lockfile_cmd = Command::new(&runner.target_compiler.cargo);
+        lockfile_cmd.arg("generate-lockfile");
+        lockfile_cmd.arg("--manifest-path").arg(STDARCH_RUST_PROGRAMS.manifest_path(&runner.dirs));
+        spawn_and_wait(lockfile_cmd);
```
bjorn3, line 377: Please instead put the lockfile in the patches directory of this repo and copy it over.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1667/c232e6289b7757783926a9970618814948f9457d..180f2122291155a0d90f0293acf33c4c2a448b22)*

bjorn3 (changes requested):

build_system/tests.rs
CathalMullan: Seems to be around the same locally. I'll run some CI tests to see if there's any meaningful difference.

build_system/tests.rs
CathalMullan: This is the lockfile of the generated workspace, which I assumed would change too often. Maybe it's stable enough to manage via a patch.

Mentions generated packages like:

```toml
[[package]]
name = "mod_9"
version = "0.1.0"
dependencies = [
 "cc",
 "core_arch",
]
```

build_system/tests.rs
CathalMullan: Runs slightly faster in CI without `--release`, I'll drop it.

build_system/tests.rs
CathalMullan: Right, they generate a `mod` per CPU core. So not sure a patch is feasible here.

https://github.com/rust-lang/stdarch/blob/80ade55ace1388cbb634f06174f6c0813747250b/crates/intrinsic-test/src/common/mod.rs#L126-L131

bjorn3: Since https://github.com/rust-lang/stdarch/pull/2202 intrinsic-test generates a lockfile. Can you rebase and remove the `cargo generate-lockfile`?

CathalMullan:
Rebased and dropped the lockfile generation.

Refreshed the skip lists as well.
Hit a `clang` specific failure, so added per `$CC` skipping too.

The patch approach also needed to be tweaked, now filtering out SVE/SVE2 generated tests manually, due to https://github.com/rust-lang/stdarch/pull/2160.

bjorn3:
Thanks a lot for implementing this!

Issue #1664, opened 2026-06-07
ICE `expected abort due to worker thread errors`


matthiaskrgr:
snippet:
````rust
//@compile-flags: -Zcodegen-backend=cranelift
const SZ: usize = 64_000_000;
type BigDrop = [String; SZ];
fn f2(_moveme: BigDrop) -> String {
    let [a, ..] = _moveme;
    a
}
fn main() {
    f2(std::array::from_fn(|_| String::new()));
}

````

Version information
````
rustc 1.98.0-nightly (43a4909ee 2026-06-07)
binary: rustc
commit-hash: 43a4909ee98ed4d006d9d773f5d94dc58e34f846
commit-date: 2026-06-07
host: x86_64-unknown-linux-gnu
release: 1.98.0-nightly
LLVM version: 22.1.6

````

Possibly related line of code:
https://github.com/rust-lang/rust/blob/43a4909ee98ed4d006d9d773f5d94dc58e34f846/compiler/rustc_codegen_ssa/src/back/write.rs#L2107-L2119

Command:
`/home/matthias/.rustup/toolchains/master/bin/rustc -Zcodegen-backend=cranelift`

<details><summary><strong>Program output</strong></summary>
<p>

```
error: backend implementation limit exceeded while compiling _RINvNtCslXUzWlqvauF_4core5array11try_from_fnINtNtNtB4_3ops9try_trait17NeverShortCircuitNtNtCskgHWY1ejRYb_5alloc6string6StringEKj3d09000_NCINvMBJ_BG_10wrap_mut_1jNCNvCs15CMijXIseV_4mvce4main0E0EB2D_


thread 'rustc' (3850296) panicked at /rustc-dev/43a4909ee98ed4d006d9d773f5d94dc58e34f846/compiler/rustc_codegen_ssa/src/back/write.rs:2113:17:
expected abort due to worker thread errors
stack backtrace:
   0:     0x7f6189092c59 - <<std[94ce253409a2fa44]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[ffdaff015e106f9f]::fmt::Display>::fmt
   1:     0x7f6189822b88 - core[ffdaff015e106f9f]::fmt::write
   2:     0x7f61890a8ce6 - <std[94ce253409a2fa44]::sys::stdio::unix::Stderr as std[94ce253409a2fa44]::io::Write>::write_fmt
[...]

Issue #1656, opened 2026-05-28
ICE `guard_size * probe_count` is too large to fit in a 32-bit immediate


matthiaskrgr:
might be similar to https://github.com/rust-lang/rustc_codegen_cranelift/issues/1552

snippet:
````rust
//@compile-flags: -Zcodegen-backend=cranelift
fn main() {
    assert_eq!(&mut [0; 536870787_usize][..], &mut []);
}

````

Version information
````
rustc 1.98.0-nightly (8c1d72083 2026-05-28)
binary: rustc
commit-hash: 8c1d72083bb2b7b9df8cbbf06a57d240a007bd85
commit-date: 2026-05-28
host: x86_64-unknown-linux-gnu
release: 1.98.0-nightly
LLVM version: 22.1.6

````

Command:
`/home/matthias/.rustup/toolchains/master/bin/rustc -Zcodegen-backend=cranelift`

<details><summary><strong>Program output</strong></summary>
<p>

```

thread '<unnamed>' (3650454) panicked at /rust/deps/cranelift-codegen-0.131.0/src/isa/x64/inst/emit.rs:373:18:
`guard_size * probe_count` is too large to fit in a 32-bit immediate: TryFromIntError(PosOverflow)
stack backtrace:
   0:     0x7fa36dcb8189 - <<std[d49b81100c593501]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[307308a59b42f609]::fmt::Display>::fmt
   1:     0x7fa36e420988 - core[307308a59b42f609]::fmt::write
   2:     0x7fa36dccee76 - <std[d49b81100c593501]::sys::stdio::unix::Stderr as std[d49b81100c593501]::io::Write>::write_fmt
   3:     0x7fa36dc8e3ae - std[d49b81100c593501]::panicking::default_hook::{closure#0}
   4:     0x7fa36dcabac3 - std[d49b81100c593501]::panicking::default_hook
   5:     0x7fa36caa7ff1 - std[d49b81100c593501]::panicking::update_hook::<alloc[41b0df45fb608239]::boxed::Box<rustc_driver_impl[98011a5ec9f8d80f]::install_ice_hook::{closure#1}>>::{closure#0}
   6:     0x7fa36dcabda2 - std[d49b81100c593501]::panicking::panic_with_hook
   7:     0x7fa36dc8e472 - std[d49b81100c593501]::panicking::panic_handler::{closure#0}
[...]

PR #1669, opened 2026-07-01
Report CodegenError::Unsupported as a rustc diagnostic instead of panicking

soumik15630m:
## Summary

When `Context::compile` returns `Err(CodegenError::Unsupported(_))`,
`src/base.rs` currently falls through to the generic catch-all arm and
panics:

```rust
Err(err) => {
    panic!("Error while defining {name}: {err:?}", name = codegened_func.symbol_name);
}
```

This turns a recoverable, known-cause compilation error into an ICE.
The codebase already has a precedent for handling a specific
`CodegenError` variant gracefully — `CodegenError::Verifier` is caught
in its own arm just above and reported via `early_dcx.early_fatal`
instead of panicking. This PR adds a matching arm for
`CodegenError::Unsupported`.

## Motivation

This gap was found while working on
bytecodealliance/wasmtime#13783, which changes x64 codegen to return
`CodegenError::Unsupported` instead of panicking when a function's
stack frame is too large to encode (see
rust-lang/rustc_codegen_cranelift#1656). Before that PR, Cranelift
never returned this error variant from this call site, so this gap in
cg_clif's own error handling was never exercised. With rust-lang/rustc_codegen_cranelift#13783's fix
applied, `Context::compile` now reaches this code path and previously
hit the generic `panic!`, producing an ICE with a stack trace instead
of a clean compiler error.

## Testing

Built this against a local patch of bytecodealliance/wasmtime#13783
(via `[patch.crates-io]`) and rebuilt with `y.sh build`. Compiling the
repro from rust-lang/rustc_codegen_cranelift#1656
(`let x = [0; 536870787_usize];`) now produces:error: cranelift codegen error: stack frame size exceeds the 2GB limit supported by the x64 backend (see rust-lang/rustc_codegen_cranelift#1656)

instead of panicking on the `Err` at this call site.

## Known follow-up (not fixed here)

With this change, the error is reported cleanly, but the worker thread
that calls `early_fatal` from inside this closure then triggers a
secondary panic in `rustc_codegen_ssa` ("expected abort due to worker
thread errors", `back/write.rs:2107`). This reproduces for the
*existing* `Verifier` arm too, using this same `early_fatal` pattern —
it's a pre-existing issue in how rustc's codegen coordinator handles a
worker thread calling `early_fatal`, tracked separately in rust-lang/rustc_codegen_cranelift#1664, and
out of scope for this PR.

### DIFF
--- a/src/base.rs
+++ b/src/base.rs
@@ -220,6 +220,12 @@ pub(crate) fn compile_fn(
                 );
                 early_dcx.early_fatal(format!("cranelift verify error:\n{}", pretty_error));
             }
+            Err(ModuleError::Compilation(CodegenError::Unsupported(msg))) => {
+                let early_dcx = rustc_session::EarlyDiagCtxt::new(
+                    rustc_session::config::ErrorOutputType::default(),
+                );
+                early_dcx.early_fatal(format!("cranelift codegen error: {msg}"));
+            }
             Err(err) => {
                 panic!("Error while defining {name}: {err:?}", name = codegened_func.symbol_name);
             }


bjorn3:
CodegenError::Unsupported should only happen when cg_clif uses an instruction or feature not supported by a Cranelift backend. This would be a bug in cg_clif and thus should be a panic rather than error message.

soumik15630m:
sure

PR #1674, opened 2026-07-11
remove f16-math patch

folkertdev:
We now have fallbacks via f32 for the intrinsics in core since https://github.com/rust-lang/rust/pull/150946.

cc https://github.com/rust-lang/rust/pull/158972


bjorn3:
Can 0001-portable-simd-Disable-f16-usage-in-portable-simd.patch be removed now too? And would you mind rebasing again? I had to change the libcall code to fix the abi for f128 on s390x.

folkertdev:
Just rebased

I believe `0001-portable-simd-Disable-f16-usage-in-portable-simd.patch` is specifically needed on windows, and the issue it runs into is fixed in GCC 16 but our CI does not use that yet. I can try removing it in a separate PR though to clarify.

PR #1676, opened 2026-07-16
use cranelift's fma for `fmuladdf`

folkertdev:


src/intrinsics/mod.rs
```diff
@@ -367,12 +375,6 @@ fn codegen_float_intrinsic_call<'tcx>(
         sym::fmaf64 => ("fma", 3, fx.tcx.types.f64, types::F64),
         sym::fmaf128 => ("fmaf128", 3, fx.tcx.types.f128, types::F128),
```
folkertdev, line 376: so why don't we also want this for the 4 intrinsics here? It looks to me like `fmuladd` is not really hooked up to anything at all in `core`, it just exists as an intrinsic?

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1676/1ca121740cf4fcda586e1ce665962f8b936e790b..d6c5fbcdbd3182a8a9493249b39dabf647212ae2)*

PR #1678, opened 2026-07-23
Rustc pull update

Kobzol:
Merge ref '390279b302ca' from rust-lang/rust

Pull recent changes from https://github.com/rust-lang/rust via Josh.

Upstream ref: rust-lang/rust@390279b302ca98ae270f434100ae3730531d1246
Filtered ref: rust-lang/rustc_codegen_cranelift@c70baf14407e0c1fcd8df30b80d499e81de44435
Upstream diff: https://github.com/rust-lang/rust/compare/...390279b302ca98ae270f434100ae3730531d1246

This merge was created using https://github.com/rust-lang/josh-sync.


josh-sync.toml
```diff
@@ -0,0 +1,3 @@
+repo = "rustc_codegen_cranelift"
+filter = ":rev(<=cb89537d435ec9cc9c812cd47c6130bd96aa0050:prefix=compiler/rustc_codegen_cranelift):/compiler/rustc_codegen_cranelift"
```
christian-schilling, line 2: This needs to be
`:~(history="keep-trivial-merges")[:rev(<=cb89537d435ec9cc9c812cd47c6130bd96aa0050:prefix=compiler/rustc_codegen_cranelift)]:/compiler/rustc_codegen_cranelift`

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1678/1ae2edd4e2cea2fec48e9a936817794f111df719..d3c5613331a2d940c7b6a1cf44d0390da135dbeb)*

josh-sync.toml
```diff
@@ -0,0 +1,3 @@
...
+sync-version = 2
```
RalfJung, line 3: Is that a flag that exists in josh-sync? I can't find it...

I would have suggested to rename it since I don;t know what "version of the sync" is supposed to mean. I would have called it `josh-sync-config-version` or so.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1678/1ae2edd4e2cea2fec48e9a936817794f111df719..d3c5613331a2d940c7b6a1cf44d0390da135dbeb)*

Kobzol: It's still WIP: https://github.com/rust-lang/josh-sync/pull/54

The config itself is backwards compatible, so it's not a new major version of the config, per-se. It is a new way to perform synces (with a different filter, essentially). Which is why I named it `sync-version`.

It could be also named `filter-version`, I guess, that is a bit more accurate.

RalfJung: I like `filter-version`. :)

PR #1679, opened 2026-07-23
Rustc pull update

Kobzol:
Merge ref '165cce8d820b' from rust-lang/rust

Pull recent changes from https://github.com/rust-lang/rust via Josh.

Upstream ref: rust-lang/rust@165cce8d820b229af8f6a8226cf0b910b57600ff
Filtered ref: rust-lang/rustc_codegen_cranelift@c70baf14407e0c1fcd8df30b80d499e81de44435
Upstream diff: https://github.com/rust-lang/rust/compare/...165cce8d820b229af8f6a8226cf0b910b57600ff

This merge was created using https://github.com/rust-lang/josh-sync.


example/mini_core.rs
```diff
@@ -6,6 +6,7 @@
     extern_types,
     decl_macro,
     rustc_attrs,
+    rustc_private,
```
bjorn3: This repo and the rust-lang/rust repo got desynced I guess. CI fails with this feature enabled.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rustc_codegen_cranelift/1679/1ae2edd4e2cea2fec48e9a936817794f111df719..5a2258b34b858e8fb548832dd353a31c6fffdbca)*

Kobzol: This change was created in the pull PR even if I tried it a ~month ago, I'm not sure where is it coming from. Should I just remove this change? Not sure if CI will be green in rust-lang/rust then, though.

Kobzol:
Ok, CI is green.

It would be great if someone else could independently reproduce
[#t-infra > rustc_codegen_cranelift Josh integration @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/rustc_codegen_cranelift.20Josh.20integration/near/612402397) on their fork though, to make sure that it also works elsewhere, before we land this.

Issue #1686, opened 2026-08-07
Bench CI job fails since latest nightly


bjorn3:
https://github.com/rust-lang/rustc_codegen_cranelift/actions/runs/31185068027/job/92898490952

I can't reproduce this failure locally. I can reproduce the test_llvm job failure though. That one is https://github.com/rust-lang/portable-simd/issues/546

PR #1687, opened 2026-08-07
Prevent `RUSTFLAGS` leak in bench build

CathalMullan:
Fixes one half of rust-lang/rustc_codegen_cranelift#1686

Job was failing due to a deprecated error:

```
error: use of deprecated constant `std::f64::INFINITY`: replaced by the `INFINITY` associated constant on `f64`
  --> src/ray.rs:24:38
   |
24 |         let mut distance = std::f64::INFINITY;
   |                                      ^^^^^^^^
   |
   = note: `-D deprecated` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(deprecated)]`
```

### DIFF
--- a/build_system/bench.rs
+++ b/build_system/bench.rs
@@ -67,7 +67,7 @@ pub(crate) fn benchmark(dirs: &Dirs, compiler: &Compiler) {
         target_dir = target_dir.display(),
     );
     let llvm_build_cmd = format!(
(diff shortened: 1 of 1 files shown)


bjorn3:
That explains why I couldn't reproduce it locally. Thanks a lot for figuring this out!

